PR #1, opened 2021-01-24
Set up Docker dev environment

dfabulich:
One of the goals of using Docker is to allow us to use a dev environment that’s similar to prod, but I don’t feel that the dev environment I have here is even close to what we’d want in prod. Still, it's such a hassle to manually set up Apache and MySQL that I think this will be a huge improvement.


Durafen:
Works really well, was been able to start using the dev environment in less than 5 min.

PR #2, opened 2021-01-24
Load boxart images from production IFDB

dfabulich:


dfabulich:
I rolled this into PR #2

PR #10, opened 2021-01-25
Switch to MariaDB

dfabulich:
This PR switches the IFDB dev environment to MariaDB 10. It seemed to work when I smoke tested it.

Note that IFDB does _not_ work with the current latest MySQL 8, which uses ICU regular expressions instead of MySQL 5 "Henry Spencer" regexes. https://dev.mysql.com/doc/refman/8.0/en/regexp.html#regexp-compatibility

For example, if you just change the `mysql` line in `docker-compose.yml` to `mysql:8`, searches by author fail `http://localhost:8080/search?searchfor=author%3AEmily+Short` because the query includes this "Henry Spencer" regex: `author like '%Emily Short%'  or author rlike '[[:<:]]E.*[[:space:]]+Short[[:>:]]'`

I did confirm that search by author works as expected in MariaDB 10, but I think without a clear test plan, it's hard to be sure whether this is safe.


dfabulich:
Ah, I found another bug. Switching from MySQL 5 to MariaDB 10 introduces character encoding bugs, e.g. http://localhost:8080/viewlist?id=k7rrytlz3wihmx2o `Recommendations by Pegbiter (MalmÃ¶, Sweden)` which should say `Malmö, Sweden`

dfabulich:
I think I've fixed the character encoding issues. MariaDB was defaulting to UTF-8 character encoding. I don't know why … according to this page https://mariadb.com/kb/en/server-system-variables/#character_set_server the default is `latin1` (ISO-8859-1), which is what IFDB wants/expects.

But by forcing MariaDB to use latin1, the character encoding bugs seem to have fixed themselves.

Issue #8, opened 2021-01-25
Sql table clubs missing 2 columns


Durafen:
Accessing any club page results in error: 
```
The requested club is not in the database. It might have been deleted, or the link you followed to reach this page might be broken.
```
This happens because the query tries to access 2 columns (`password` and  `pswsalt`) that doesn't exist in the current sql dump.
In the current state, those 2 columns can added with empty data (it fixes the error).

dfabulich:
Thanks! I constructed the `patch-schema.sql` script by hand, and I somehow missed these columns. I double checked it this time; I'm pretty sure I've got them all now.

PR #13, opened 2021-01-25
Club - Fix error while joining club

Durafen:
There was a missing parameter for the db in mysql_query command.

Follow up and a fix to [Issue #8: Error when trying to join a club](https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/8)

### DIFF
--- a/www/club
+++ b/www/club
@@ -298,7 +298,7 @@ if (isset($_REQUEST['join']))
 
     // get the current user's wish list privacy
     $result = mysql_query(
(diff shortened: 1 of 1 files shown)


dfabulich:
Thanks! I'd never used "clubs" in IFDB before. I wonder if this was broken on ifdb.tads.org as well…?

Durafen:
I think it was broken for ages.
The first time I saw this error on ifdb.tads.org was about 2 years ago.
Now I can sleep in peace  :)

PR #14, opened 2021-01-25
newitems - removing unnecessary br that affects allnew page

Durafen:
Leftovers from debugging MySQL errors printed `<br>` tag and affected `allnew` page.

Original issue : [The allnew page has unicode issues ](https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/7)

### DIFF
--- a/www/newitems.php
+++ b/www/newitems.php
@@ -43,7 +43,6 @@ function getNewItems($db, $limit)
          from games
          order by created desc
          $limit", $db);
-    echo mysql_error($db)."<br>";
     $gamecnt = mysql_num_rows($result);
     for ($i = 0 ; $i < $gamecnt ; $i++) {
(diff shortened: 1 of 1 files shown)


dfabulich:
That `<br>` is a bit embarrassing, but I claim that it's not actually the cause of https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/7

For example, when I navigate to http://localhost:8080/allnew I see "It doesn’t matter what they say..." on my local dev box, with a curly apostrophe in "doesn’t." But when I go to https://ifdb.tads.org/allnew?&pg=2 and scroll down to the corresponding review, there's a unicode error, "It doesn�t matter what they say..."

The `<br>` at the top does introduce a single line of whitespace at the top of the page, so I'm more than happy to merge this just to make that go away! Hopefully the UTF-8 error will go away on its own when we deploy the site.

PR #17, opened 2021-01-26
Change domain name

### LINKED ISSUE #7: Replace all ifdb.tads.org with ifdb.org
As we do the server transition, we'll want to ensure that we don't link to ifdb.tads.org.

I see a bunch of links to ifdb.tads.org in the database archive, and also in the code. (We especially don't want to link to ifdb.tads.org in emails we send out, e.g. new-user emails.)

dfabulich:
Working on #7.

This is a tough one to review, especially since we don't have a test plan. I tried to break it up into several commits.

1. Update email address: We'll always use the same domain for this, ifdb.org. The code should never suggest emailing dev.ifdb.org.
2. HTTPS redirect: I _think_ I set it up correctly so it will redirect from HTTP to HTTPS in staging and production but not in local dev. All I've actually tested is that it doesn't redirect in local dev.
3. xmlns: I'm not entirely sure if I should change these at all. (But it can't be right to just leave these URLs pointing to ifdb.tads.org, right?) I _just_ changed the domain names without switching to `https://` but it's not at all clear to me that this is the right call.
4. This is the big one, but it's also the simplest: I introduce a `get_root_url()` utility in `utils.php` and then use it all over the place instead of a hardcoded string. (It's plausible that some of these _should_ be hard coded strings; more eyes are very welcome here.)
5. Colophon: Here's a weird one. It's about the `viewgame` API. https://ifdb.tads.org/api/viewgame 
    > The viewgame API lets applications retrieve IFDB's listing data for a game in the Treaty of Babel's iFiction XML format.
    
    The "Treaty of Babel" has a `<colophon>` section with a `<generator>` element. I'm hard coding this to `ifdb.org/viewgame`. (Right??)
6. IFArchive pending URLs: These are actually stored in the database, so we don't want them to vary depending on whether you're in local dev/staging/production.
7. OpenSearch descriptor: This is what allows your browser to help you search IFDB straight from the URL bar. I suppose I _could_ generate the OpenSearch descriptor via PHP and use `get_root_url()`, but, meh.
8. Dev images: Lastly, I undid part of my change by setting the production URL back to ifdb.tads.org, because for now that's the only place for local developers to get their images. Once we're deployed to prod for real, we'll change this setting and then developers can get their images from prod. (Or _maybe_ by then we'll have a way of storing/retrieving the images directly from IFArchive.)
9. Data migration: We're running a basic regex on the SQL as we import it to replace all ifdb.tads.org references in the SQL to ifdb.org

### DIFF
--- a/prepare_dev_environment.sh
+++ b/prepare_dev_environment.sh
@@ -9,7 +9,8 @@ unzip -o $FILENAME
 
 rm -rf initdb
 mkdir initdb
-cat create_db.sql ifdb-archive.sql > initdb/00-init.sql
+cp create_db.sql initdb/00-init.sql
+perl -pe 's!https?://ifdb.tads.org!https://ifdb.org!g' < ifdb-archive.sql >> initdb/00-init.sql
 cp patch-schema.sql initdb/01-patch-schema.sql
 cp create-admin.sql initdb/02-create-admin.sql
 
--- a/prepare_full_dev_environment.sh
+++ b/prepare_full_dev_environment.sh
@@ -5,8 +5,8 @@ IFDB_FULL_EXPORT=$1
 rm -rf initdb
 mkdir initdb
 echo "create database if not exists ifdb; use ifdb;\n" > initdb/init.sql
-# remove bogus definers to fix views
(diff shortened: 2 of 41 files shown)


vaughany (approved):
Looks good to me. :+1:

PR #23, opened 2021-01-27
Add the option to flag a review as inappropriate

dfabulich:
The site already had a "Flag spoiler" feature. This adds a "Flag as inappropriate" feature.

It doesn't do much; it records the flag in the `reviewflags` table and it emails `ifdbadmin`. But hey, it's a start.

<hr>

Here's where you'd find the feature
image

<hr>

Here's the submission screen
image

<hr>

The resulting email looks like this. You can reply to the email to reply to the flagger, or click through to the review to take action there.
image


Durafen (approved):
Works, flagged and email sent.
Imho it will be enough for a while.

PR #25, opened 2021-01-29
Add delete and edit reviews and comments to admin

Durafen:
I have done testing, but more eyes are always welcome.

Closes [Issue #26](https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/26)

### DIFF
--- a/www/review
+++ b/www/review
@@ -25,8 +25,10 @@ $isOldReview = false;
 // get the game ID
 $id = $_REQUEST['id'];
 
-if ($_REQUEST['userid']) $userid = $_REQUEST['userid'];
-else $userid = $_SESSION['logged_in_as'];
+$userid = $_SESSION['logged_in_as'];
+if ($_REQUEST['userid'] && check_admin_privileges($db, $userid)) {
+    $userid = $_REQUEST['userid'];
+}
 
 // ------------------------------------------------------------------------
 //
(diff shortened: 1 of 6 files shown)


www/review
```diff
@@ -26,6 +25,8 @@ $isOldReview = false;
 // get the game ID
 $id = $_REQUEST['id'];
 
+if ($_REQUEST['userid']) $userid = $_REQUEST['userid'];
+else $userid = $_SESSION['logged_in_as'];
```
dfabulich, line 29: Wait a minute! This isn't OK. The `$_REQUEST` variable is under the user's control. They can type anything they want in the `&userid` parameter.

For example, in your branch, login as a non-admin user and go here: http://localhost:8080/review?id=ckdsz92q8cws7r7c&userid=1k0mok394bbghxdp

Expected: You're not Jacq, so you shouldn't be allowed to edit her review
Actual: You're not only allowed to see the "edit review" page, you can edit the review, save, and publish it, as the wrong user.

dfabulich: I think it should be OK if we do this (but I haven't tested it):

```php
$userid = $_SESSION['logged_in_as'];
if ($_REQUEST['userid'] && check_admin_privileges($db, $userid)) {
    $userid = $_REQUEST['userid'];
}
```

dfabulich (changes requested):
I think I caught a security issue here!

www/review
Durafen: Updated the code.
Nice catch, thanks.

PR #26, opened 2021-02-06
Update README.md

qdacsvx:
Getting started: add instruction to clone the repository.


dfabulich:
Different users use different URLs to clone the repository (HTTPS vs SSH vs Github CLI).

Indeed, you don't even have to use git to clone it; if you click the Code you can use the "Download ZIP" feature. You just gotta get the files somehow; I think it's OK to leave "getting the files" up to the reader.

PR #28, opened 2021-02-18
Tags removal for admins

Durafen:
This is a first working draft for tags removal for admin users.
Please check it, I will also make more tests in the following days.

### DIFF
--- a/www/taggamedelete
+++ b/www/taggamedelete
@@ -1,0 +1,145 @@
+<?php
+
+@session_start();
+include_once "util.php";
+include "pagetpl.php";
+include "login-check.php";
+
+function sendResponse($statmsg, $errmsg, $detail, $tagInfo)
+{
+    global $xml;
+    
+    if ($xml)
+    {
+        header("Content-Type: text/xml");
+        header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
+        header("Cache-Control: no-store, no-cache, must-revalidate");
+
+        echo "<?xml version=\"1.0\" encoding=\"UTF-8\" ?><response>"
+            . ($statmsg ? "<label>$statmsg</label>" : "")
+            . ($errmsg ? "<error>$errmsg</error>" : "")
+            . $tagInfo
+            . "</response>";
+    }
+    else
+    {
+        smallPageHeader("Tag a Game");
+        if ($errmsg)
+            echo "<span class=errmsg>$errmsg</span><p>";
+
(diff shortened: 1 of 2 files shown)


dfabulich:
I noticed a few issues.

First issue: Cancel button doesn't work on "delete tags" window

To reproduce: Sign in as admin, navigate to a game with tags, click on "Delete" to open the "Delete" window, and click on an "x" to delete a tag. But now, you change your mind, and click the "Cancel" button in the upper right.

Expected: Since you cancelled the changes, none of the tags should be deleted.
Actual: The tag was deleted as soon as you clicked the "x." If you refresh the game details page, the tag is gone.

Second issue: Clicking on a tag name in the "delete tags" window doesn't delete a tag

To reproduce: Sign in as admin, navigate to a game with tags, and click on "Delete" to open the delete window. Instead of clicking on the "x" icon, click on the name of the tag, e.g. "Inform 7."

Expected: This is the Delete Tags window, so when you click on a tag, the tag should go away.
Actual: Nothing happens when you click on the tag name

Third issue: The "Delete Tags" window looks too much like the Edit Tags window

This is a screenshot of the window that pops up when you click "Delete" on the game "Catventure" (the first game on the home page in the Jan 17 IFArchive DB export).

image

* The title says "Edit Tags" but it should say "Delete Tags"
* "Tags you added are shown below with checkmarks. To remove one of your tags, simply un-check it." This message is not true of the "Delete Tags" menu.
* "Enter new tags here (use commas to separate tags):" that whole little form should be removed, I think

Durafen:
Thanks for reviewing my code.

> First issue: Cancel button doesn't work on "delete tags" window
For now, I implemented it as a simple implementation where the tag deleted immediately.
I'm going to try to do it with queue, so "cancel" button will be useful.

> Second issue: Clicking on a tag name in the "delete tags" window doesn't delete a tag
This one I made on purpose, so there will be no deletion by mistake (probably because of the previous issue).
Will revisit it after I'm done with the first issue.

> Third issue: The "Delete Tags" window looks too much like the Edit Tags window
This will bring more code, ohhh well, there are already a lot of code there :)

Durafen:
Updated, incorporated all feedback.

Durafen:
If you have any ideas for the explanatory line inside the delete dialog, I can add it.

dfabulich:
Looks good! I don't think the dialog needs explanatory text at this point.

PR #33, opened 2021-02-23
Log out banned users, polling DB frequently

dfabulich:
MJR's implementation of `logged_in`/`checkPersistentLogin` would trust the session, but if the user is signed in, the session has no way of knowing whether it's banned, so we have to poll the DB on every request.

Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/23.


Durafen (commented):
Works.
Haven't noticed any significant speed impact.

PR #38, opened 2021-02-26
Ban users without deleting them

dfabulich:
When banning a user, sandbox them, don't delete them. (This allows us to reverse a ban if we decide to do so.)

Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/22


Durafen (approved):
Have tested with 3 merges together: ban-logout-slow, mod-links and ban-without-delete.
Looks like it's working.

PR #41, opened 2021-03-07
Make capitalization consistent

alice-blue:
The other headings on the page have all major words capitalized. Also, "Staff" doesn't need to be capitalized.


dfabulich:
Thank you!

Issue #40, opened 2021-03-06
Moderator email links shouldn't automatically perform their actions


dfabulich:
The moderator emails ("new user registration" "profile change" "hyperlink review") contain links that directly perform actions in them. For example, the "new user registration" email includes links to approve, trust, or ban. All you have to do is click the link, and boom, the user is approved, trusted, or banned.

That's bad. The email link should take you to the user profile page, instead, where you can take the action there with an additional click.

**Worst of all, the moderator email links don't require login.** The links include randomized "nonce" temporary passwords, allowing anyone with the link to securely perform moderator actions even if/when they're not logged in. For example, here's an old link to ban a spammer ("Veronamtt," the example user from our playbook).

```
https://ifdb.org/userconfirm?nonce=ba5ca08b913fd9a4ad2d4fa23cc54d60&userid=jkz7zifwqfgmax63&reviewProfile=B
```

Anyone who clicks on that link, logged in or not, can use the "nonce" to ban Veronamtt. (Or approve Veronamt;, or _trust_ Veronamtt.)

This might seem like an _enormous_ security issue, but it may not be _quite_ as serious as it looks, because the nonce passwords are only supposed to be emailed to trusted moderators, and each nonce expires pretty quickly.

Still, it's bad. I'm sure this was a very convenient system for MJR, but I'm gonna turn it off.

PR #42, opened 2021-03-07
Remove nonce links from moderator emails

dfabulich:
Fixes issue #40.

It's probably easiest to review each commit one at a time.

My approach to testing was as follows:

1. In an incognito window, I created a new user. I checked the email and clicked on links.
2. Then I clicked the activation link to activate the user, logged in as the new user, and added a link to my profile. I checked the "profile change" email and clicked on links.
3. As the new user, I posted a test review containing a link to https://example.com. I checked the "hyperlink review" email and clicked on links.
4. As the new user, I posted a test comment (a response to a review) containing a link to https://example.com. I checked the "comment spam flag" email and clicked on links.

I ran those tests locally on `localhost:8080` and then I deployed the code to dev.ifdb.org and ran the tests there, too, which resulted in real live emails appearing in my inbox.

One core question in my mind was: did I miss an email? I don't think so; I grepped the code for "ifdbadmin" and I think I got them all.

I also grepped for "nonce." The only remaining instances are:

* a comment in `showuser`
* `util.php`, which still has a `create_nonce` function and a `validate_nonce` function
* the `userconfirm` page, which still knows how to validate and use nonces; we just never call `create_nonce` any more


Durafen (approved):
Looking good, have tested these options.
If an email was missed, we will find it out in the next days.

dfabulich:
thanks!

PR #44, opened 2021-03-07
Update help-review-votes

alice-blue:
Add a sentence from the community moderation page https://ifdb.org/help-comm-mod. (Now that we have mods, most of the information on the community moderation page is unnecessary, so if we add this relevant part here, I think we won't need the community moderation page at all.) Add explanation of "Flag as inappropriate" option.

Relates to https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/166

### DIFF
--- a/www/help-review-votes
+++ b/www/help-review-votes
@@ -9,8 +9,6 @@ helpPageHeader("Review Voting Options");
 
 <p>These options let you "review the reviewer," to improve your own
 experience with IFDB and help guide other members to the best reviews.
-They make up IFDB's <a href="help-comm-mod">community moderation</a>
-system for game reviews.
 
 <p><b>Helpful (Yes/No):</b> Vote Yes if you think the review does a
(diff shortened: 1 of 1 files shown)


www/help-review-votes
```diff
@@ -49,6 +50,11 @@ responsibly.  <b>Don't</b> flag a review for spoilers that are already
 properly hidden with "click to see" warnings - only flag it if the
 spoilers are out in the open.
 
+<p><b>Flag as inappropriate:</b> Notify the moderators that this 
+review may violate the 
+<a href="https://ifdb.org/code-of-conduct">Code of Conduct</a> or the 
+<a href="https://ifdb.org/tos">Terms of Service</a>.
```
dfabulich, line 56: Generally we should use relative links, like `<a href="/code-of-conduct">` instead of absolute links to ifdb.org, so the links will DTRT when testing on local machines or on the dev site.

(I'll fix it.)

PR #49, opened 2021-03-09
Highlight site news in /allnew and home page

dfabulich:
In this PR, site news now appears on the IFDB home page and on the `/allnew` page.

Additionally, if the latest ("freshest") site news is less than one week old, we sort it to the top of the list.

To test this, I logged in as admin, created site news, and then posted a review. The site news sorted to the top of the list, even though the review came second.

To double check, I went into the DB and ran this query:

```sql
update sitenews set posted='2021-03-05' where itemid=77;
```

Where `77` was the itemid of my new site news item. I could see that the site news from days ago appeared above the review from today. Then I set the sitenews date back to `2021-01-17`. The site news returned to its natural sort position in the list.

image

### DIFF
--- a/www/newitems.php
+++ b/www/newitems.php
@@ -40,7 +40,8 @@ function getNewItems($db, $limit)
         "select
            itemid as sitenewsid, title, ldesc as `desc`,
            posted as d,
-           date_format(posted, '%M %e, %Y') as fmtdate
+           date_format(posted, '%M %e, %Y') as fmtdate,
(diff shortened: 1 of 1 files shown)


Durafen (commented):
Looking good, very nice addition.

`<div class="site-news">` doesn't have a closing tag, so it's contains all the inner objects in it.
If easier to spot if you add 2 news items, you can see they are close to each other, and every one of them opening new site-news which contains all the rest objects and news.

PR #51, opened 2021-03-09
Update default list of systems

dfabulich:
The "systems" field is just a text box, so you can type in anything you want, but we provide a dropdown of suggestions. Currently the dropdown provides a lot of archaic suggestions, and is missing a number of popular modern suggestions.

I ran this DB query to get a sense of what should appear on the list; it's a count of games by system.

```
MariaDB [ifdb]> select count(id), system from games where now() < date_add(published, interval 3 year) group by system having count(id) > 4 order by count(id) desc;
+-----------+-------------------+
| count(id) | system            |
+-----------+-------------------+
|       492 | Twine             |
|       173 | Inform 7          |
|        84 | NULL              |
|        83 | Choicescript      |
|        54 | Custom            |
|        45 | Adventuron        |
|        38 | StoryNexus        |
|        27 | ChooseYourStory   |
|        25 | Ink               |
|        24 |   ChooseYourStory |
|        18 | Unity             |
|        15 | Quest             |
|        15 | ADRIFT            |
|        13 | None              |
|         9 | ZIL               |
|         9 | Ren'Py            |
|         9 | TADS 3            |
|         8 | Inform 6          |
|         5 | ADRIFT 5          |
|         5 | Texture           |
+-----------+-------------------+
```


dfabulich:
Did I add/remove the right systems?

brirush84:
The only issue I see potentially arising is that there's a good chance someone will wonder about Hugo at some point. Not many games get made with it, but the ones that are are usually good ones, and I could see the authors getting miffed. Might not come up for a few years. I'd say it doesn't matter, but then Dialog is probably in the same boat, and we're adding it.

Texture seems like a system that was briefly popular but kind of dying down. It's worth adding as long as we periodically check this (every year or two), but if this is meant to be the list for the next 5 years may not be worth adding (although watch, they'll add more functionality to it and it'll take off lol)

Durafen (approved):
I don't know all the formats, but code-wise its looking good.

lkcampbell:
Glad to see the new additions. I suspect there is a small chance that some of the older authoring system might be used again, but the ability to enter in anything in the text box is a reasonable work around. And if someone does get miffed, it looks like adding a default back is a very easy change.

alice-blue:
This should resolve https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/172

PR #55, opened 2021-03-09
Update new user page to link to intfiction.org instead of RAIF

alice-blue:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/189


dfabulich:
Thanks!

PR #56, opened 2021-03-10
View disscussion from inbox fix

Durafen:
This will fix the bug with viewing full discussion from inbox.

### DIFF
--- a/www/commentutil.php
+++ b/www/commentutil.php
@@ -482,7 +482,7 @@ function revealPlonkedComment(ele)
         echo "<div class=reviewCommentReply>";
 
         // show the child list
-        showCommentList($commentPage, $itemAuthor, $cidx, $children);
+        showCommentList($db, $commentPage, $itemAuthor, $cidx, $children);
 
         // close the reply division
         echo "</div>";


dfabulich:
Nice! Could I get you to take a look at PR #49 while I have your attention?

dfabulich:
Also, do we know how to reproduce this bug (without impersonating a user)?

Durafen:
Sure, checking.

Durafen:
I also don't have enough messages in my inbox to check it, so impersonating is the way to go for me.
Haven't tried other ways.

dfabulich:
OK, I have a repro in local dev. It doesn't even require a bunch of messages.

1. Login as admin in your main browser window.
2. In an incognito window, create a new user, activate the account, then navigate to the built-in admin account http://localhost:8080/showuser?id=0000000000000000
2. Click the link there to "add a comment on this profile." Type "test 1," check the "private message" box, and Save Changes.
3. Back in your admin window, navigate to your inbox. The "test 1" message is there. Reply, typing "test 2"
4. Back in the incognito window, navigate to your inbox. The "test 2" message is there. Click "view discussion" on it.

Expected: You should see the "test 2" message there and be able to reply to it
Actual: You only see the "test 1" message that started the thread.

When I merge in your changes, this test passes.

PR #57, opened 2021-03-10
Adminops - confirm for user deletion and banning

Durafen:


Durafen:
The only solutions I can think is to make them as form with clickable buttons (and not just links), or add another new screen for confirming. 
Don't sure what is better here.

dfabulich:
IMO, it's OK for the reversible actions (Approve and Ban) to be single-click actions. If you screw them up, you'll just go back and fix it. But I think just for deletion, we should add a new screen for confirming.

@alice-blue had a good suggestion on the forum: https://intfiction.org/t/thoughts-on-its-still-too-easy-to-delete-a-user/49845

The confirmation page for deletion could require you to type in your admin user password, or even something like this:

> You're about to delete a user!
> To delete the user, type "DELETE" in the box below:
> `[                   ]`

dfabulich:
Oops, for some reason I thought you were gonna try to fix middle-clicking, but I think this is a good improvement in the short term, and I should have merged this a week ago!

(If I don't get to one of these promptly, feel free to reply to remind me to review!)

PR #66, opened 2021-03-11
Reset review vote option

Durafen:
Closes [No way to remove a review vote #195](https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/195)

### DIFF
--- a/www/reviews.php
+++ b/www/reviews.php
@@ -146,10 +146,16 @@ function displayReviewVote(reviewID, vote)
 {
     if (vote != null)
     {
-        document.getElementById("voteStat_" + reviewID).innerHTML =
-            "<br>(You previously voted "
-            + (vote == 'Y' ? "Yes" : "No")
-            + ")";
+	    if (vote == "R")
+	    {
+	    	document.getElementById("voteStat_" + reviewID).innerHTML = "";
+	    } 
+	    else {	    
+        	document.getElementById("voteStat_" + reviewID).innerHTML =
+       	  	   "<br>(You previously voted "
+	   	  	   + (vote == 'Y' ? "Yes" : "No")
+	   	  	   + ")";
+        }
     }
 }
 var curPopupMenu = null;
@@ -451,7 +457,10 @@ function showReview($db, $gameid, $rec, $specialNames, $optionFlags = 0)
             . "return false;\">Yes</a> &nbsp; "
             . "<a href=\"needjs\""
             . "onclick=\"javascript:sendReviewVote('$reviewid', 'N');"
-            . "return false;\">No</a> &nbsp; ";
+            . "return false;\">No</a> &nbsp; "
+            . "<a href=\"needjs\""
+            . "onclick=\"javascript:sendReviewVote('$reviewid', 'R');"
+            . "return false;\">Remove vote</a> &nbsp; ";
 
 			if (check_admin_privileges($db, $curuser)) {
 				echo "<a href=\"review?id=$gameid&userid=$userid\">Edit</a>&nbsp; ";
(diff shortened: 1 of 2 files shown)


www/reviews.php
```diff
@@ -439,6 +445,8 @@ function showReview($db, $gameid, $rec, $specialNames, $optionFlags = 0)
             $result = mysql_query(
                 "select vote from reviewvotes
                  where reviewid = '$qreviewid' and userid = '$curuser'", $db);
+                 error_log("select vote from reviewvotes
+                 where reviewid = '$qreviewid' and userid = '$curuser'");
```
dfabulich, line 449:
```suggestion
```

www/reviewvote
```diff
@@ -75,23 +75,33 @@ mysql_query("delete from reviewvotes
...
+	} else {
+	    sendResponse("Not Saved", "An error occurred updating the database. "
+	                 . "Please try again later.", false);
+	}
 } else {
-    sendResponse("Not Saved", "An error occurred updating the database. "
-                 . "Please try again later.", false);
+// Reset Vote result	
+    sendResponse("Vote Reseted - Thanks!", false,
+                 "Thanks for voting! Your vote has been reseted.", false);
```
dfabulich, line 100:
```suggestion
    sendResponse("Vote Removed - Thanks!", false,
                 "Thanks for voting! Your vote has been removed.", false);
```

"Reseted" isn't an English word (though it totally ought to be)

www/reviews.php
```diff
@@ -450,7 +458,10 @@ function showReview($db, $gameid, $rec, $specialNames, $optionFlags = 0)
             . "return false;\">Yes</a> &nbsp; "
             . "<a href=\"needjs\""
             . "onclick=\"javascript:sendReviewVote('$reviewid', 'N');"
-            . "return false;\">No</a> &nbsp; ";
+            . "return false;\">No</a> &nbsp; "
+            . "<a href=\"needjs\""
+            . "onclick=\"javascript:sendReviewVote('$reviewid', 'R');"
+            . "return false;\">Reset</a> &nbsp; ";
```
dfabulich, line 464:
```suggestion
            . "return false;\">Remove vote</a> &nbsp; ";
```

Seems like on https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/195 folks liked "Remove vote"

dfabulich (changes requested):
Nice work! A few suggestions

www/reviews.php
```diff
@@ -450,7 +456,10 @@ function showReview($db, $gameid, $rec, $specialNames, $optionFlags = 0)
             . "return false;\">Yes</a> &nbsp; "
             . "<a href=\"needjs\""
             . "onclick=\"javascript:sendReviewVote('$reviewid', 'N');"
-            . "return false;\">No</a> &nbsp; ";
+            . "return false;\">No</a> &nbsp; "
+            . "<a href=\"needjs\""
+            . "onclick=\"javascript:sendReviewVote('$reviewid', 'R');"
+            . "return false;\">Remove vote</a> &nbsp; ";
```
dfabulich, line 462: I just noticed that it _always_ includes the "Remove vote" link, even if you've never voted, and even right after you've removed your vote. (But fixing that sounds like a hassle.)

WDYT? Do you want to take a crack at making the link only show up when it's needed? Or I can just merge this now and we can file that as a separate bug?

Durafen:
Updated it, so it dynamically updates the "Remove vote" option.

dfabulich:
Thanks! (Sorry for the delay in reviewing.)

PR #67, opened 2021-03-14
Remove browser plugin

dfabulich:
This fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/202

Note that in the first commit, I deleted the plugins directory except for the opensearchdesc file, which support searching IFDB in your URL bar.

Most of those files were obviously part of the browser plugin architecture, but one of them wasn't: `search123` which gets a 500 error even today on ifdb.org. https://ifdb.org/plugins/search123 Since it doesn't work, I guess it's safe to delete…? There are no references to this page in the source that I can find.

I think the main thing to look out for is to find a regression… something that's still using something out of `plugin-detect.php` perhaps?

I also have some concern that this might somehow break Zoom for macOS, which apparently had fairly deep integration with IFDB, but I can't test Zoom for macOS, because it doesn't work on the current latest version of macOS Big Sur 11. (I have a hunch that Zoom's IFDB support was busted when we move domains from ifdb.tads.org to ifdb.org, but fixing that would presumably require someone to release a new version of Zoom, which I suppose will probably never happen now.)


Durafen (approved):
Looking good.

PR #70, opened 2021-03-15
Update off-site reviews

alice-blue:
Update examples of off-site review sources. Update urls for Xyzzy News and SPAG that are used in review headings.

Related to https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/201


dfabulich:
Thanks!

PR #71, opened 2021-03-15
Sort by starsort when sorting by "Highest Rated"

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/16

This is a tough PR.

The idea is to stop sorting games by average rating, which sorts a game with a single 5-star rating above Counterfeit Monkey (a game with more than 100 5-star ratings and some 4-star ratings), and instead sort by "starsort," which is what I'm calling Evan Miller's formula for sorting by star rating. https://www.evanmiller.org/ranking-items-with-star-ratings.html

I've deployed these changes to dev.ifdb.org. https://dev.ifdb.org/search?sortby=ratu&searchfor=&browse=1 It looks pretty good to me.

# What's a starsort?

Even with a basic familiarity with statistics, I found Miller's article _very_ difficult to read. The core idea is as follows.

EDIT: I had an older explanation here that I now realize wasn't very good. I've replaced it with a substantially clearer explanation.

There are two big ideas in starsort.

1. **We can compute a "ranking score" X by subtracting the standard deviation σ of the ratings from the average/mean M.** Given a final ranking score X = M - σ, we’re claiming that a game probably has an average rating of X or greater. The more consensus the ratings have, the narrower the standard deviation will be, i.e. the more confident we can be, and so the higher X will be.

      We also multiply the standard deviation by a constant **confidence-interval factor z**, so X = M - zσ. For example, if you want to be 90% sure that the average is greater than X, use z = 1.65; z = 1.96 represents a 95% confidence interval. Using larger z values—requiring higher certainty—will put more priority on high-consensus ratings, making it harder for games with fewer ratings to stand out. I used 1.65 in this PR.

2. **Before computing the mean, give every game five fake ratings, one per star rating.** That is, every game gets a fake 1-star rating, a fake 2-star rating, a fake 3-star rating, a fake 4-star rating, and a fake 5-star rating.

    Why? Imagine building a pie chart of probabilities for what we think the next rating will be for a game. (Statisticians call this kind of pie chart a "Dirichlet distribution.") Suppose we have a bunch of ratings at all five different star levels. The easy/natural way to build the pie chart is just to use the pie chart of the ratings themselves. So take the number of 1-star ratings, divide that by the total number of ratings, and that's the expected probability that the next rating will be a 1-star rating.

    But suppose we don't yet have any 1-star ratings. It would be foolish to predict that there's a 0% chance that the next rating will be 1-star. Or suppose we don't yet have _any_ ratings. In that case, we _should_ predict that each star rating is equally likely, a 20% 1-in-5 chance.

    When we set up the pie chart with one fake rating per level, it reflects our Bayesian prior assumptions that each star level is possible; before we have any ratings, it reflects our prior assumption that each star level is equally likely. As each data point comes in, we adjust the pie chart with posterior probabilities, but never fully delete our prior assumptions.

    As a practical matter, these five fake ratings ensure that the average rating for a game doesn't shoot up to 5-stars just because we get a single 5-star review, and, more importantly, it ensures that the standard deviation isn't 0 just because we only have one or two ratings.

Miller's article is _very_ technical, but, if you read carefully, he's just computing the mean of the <s>Dirichlet distribution</s> pie chart `f(p)`, which is literally just the mean of the ratings (including the five fake ratings), and then he subtracts the square root of the variance, i.e. the standard deviation, multiplying it by the constant factor z.

If that's not enough detail, feel free to ask me questions, or you can attempt to read and re-read Evan Miller's article, like I did.

# Reading this PR

It's easiest to review this PR commit by commit. (EDIT: This section used to be much bigger, but I rebased and rewrote the commits in a way that allows you to read the commits one-by-one.)

# Testing / Reviewing the PR

IMO the main thing to do is to make sure that my _refactors_ are safe/correct. Do any numbers on the search results change before and after those first two commits? What if you're a shadowbanned troll? Can you still see everything you would have seen before?

I should point out that someone on https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/16 suggested that we should keep around average rating. I'm… not sure that's right, but I'm not sure it's wrong, either. I'm really not fond of cluttering the drop down with extra options. Maybe there could be a link on the search results page to let you switch algorithms? But will anyone _really_ want to sort by average rating in fact??


curiousdannii:
Could the star rating be precomputed (and stored in the DB), and updated when someone adds a rating to an entry? That's likely to be many times less frequent that how often the games are sorted.

brirush84:
Something weird has happened where #ratings, rating, #reviews, and ratingdev all generate errors when used. I thought the rating one might be intentional, but having all of them not work means we're probably missing something.

brirush84:
Okay, I think it might be an issue with search123. I see commits to all the search files except that one.

dfabulich:
@brirush84 Oops! Good catch. I've rewritten this PR to fix that (and to make it much easier to review). You can once again sort by all the specials. https://dev.ifdb.org/search?searchbar=%23ratings%3A25-+rating%3A1-2

`search123` is dead code; it doesn't work, it doesn't do anything, and nobody's using it. I'm deleting it in PR #67

@curiousdannii I think it could be pregenerated, yes, but the whole view for all of IFDB generates itself in a few hundred milliseconds on my machine, so I'm not worried. Unfortunately, MySQL doesn't have a convenient way of generating materialized views, so it would be a lot of work to extract the necessary performance gains.

This PR would nonetheless be a useful first step before pregenerating the views (converting them to tables).

brirush84:
This fixes my main problems. Since I can use the 'rating' modifier to restrict actual rating, it's okay for me that the answers are sorted in starsort. I consider this overall a big improvement, as I agree that starsort is better for general browsing and this doesn't break past stuff or my own weird research sorts.

dfabulich:
I had a breakthrough just now and rewrote my description of how starsort works. It's _much_ simpler than I'd previously realized.

Durafen (approved):
Looking good.
This is a game changer, I wanted this change from my first day of using ifdb, thanks!

PR #74, opened 2021-03-16
Adopt MJR's "Midnight" styles when the browser prefers dark mode

dfabulich:
These styles were copied from https://ifdb.org/css?id=5

The images were copied from https://ifdb.org/users/asdf90813lkjf09813/css/images/

You can test it in Chrome Dev Tools, with "Emulate CSS prefers-color-scheme: dark." You can find it and turn it on and off by searching for it in the Command menu (Command+Shift+P or Control+Shift+P)

localhost_8080_


Durafen (approved):
Looking good,
We should add it to user preferences or theme chooser.

PR #75, opened 2021-03-16
Fix minor mobile bugs

dfabulich:
1. The home page prefers to put the "Vote!" and "New to IF?" sections side-by-side, but that's not a good fit on mobile. On narrow screens, we now convert these to blocks in a single column.

image

2. When viewing the site in Desktop mode, the RSS links would overlap the text. They were using absolute positioning; now we use flexbox instead.

image


Durafen (approved):
Looking good.

PR #77, opened 2021-03-16
Fix ugly searchGo parameters in search results

dfabulich:
https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/209


Durafen (approved):
Clean URL, looking good.

PR #78, opened 2021-03-16
Ensure reviews and descriptions aren't too wide

dfabulich:
On my MacBook Pro, I set my resolution pretty high in settings; IFDB displays very long lines by default. I find it hard to skim paragraphs that wide.

In this PR, I limit `<p>` tags to `60ch` wide and the main/topbar/topctl `<div>`s to `100ch`. It leaves some awkward whitespace, but the text flows a lot better IMO.

image

image

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -164,6 +164,8 @@ div.topbar {
     width: 100%;
     background: url("/ifdb-topbar.jpg") #c0c0c0 no-repeat;
     height: 55px;
+    max-width: 100ch;
+    margin: 0 auto;
 }
 a.topbar {
     height: 55px;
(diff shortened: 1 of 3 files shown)


Durafen:
This one is a little controversy, It's looking better for some pages with long texts.
IMHO I would make the change (or add URL parameter for it), and make an intfiction poll to gather users opinions.

dfabulich:
Updated with Stian's suggestion https://intfiction.org/t/wide-paragraphs/49948/4?u=dfabulich

I think this is shippable

dfabulich:
Narrowed the header, too.

Durafen (approved):
Looking good. Like this change.

Issue #11, opened 2021-01-25
Test plan


dfabulich:
We should have a test plan that ensures that everything works after we switch domains. We can also use this test plan to support major infrastructure enhancements (like PR #10) and perhaps some day as a specification for an automated test suite.

The test plan should be a set of instructions that a user could follow to manually verify that the site's major features still work:

* Navigate to the home page
* Register a new user (filling out the CAPTCHA)
* Add a new game
* ...

lkcampbell:
@dfabulich, I can take this one if you want to assign it to me. Any file format preferences? I assume we just need a text file in a `./test` directory for now, correct?

dfabulich:
That sounds fine to me! (Maybe in Markdown?)

lkcampbell:
@dfabulich, should I assume the person reading the test plan is using the local dev environment with the existing `ifdbadmin` admin user?

dfabulich:
I think that sounds right, for a start.

lkcampbell:
@dfabulich I finished the Public Home Page test plan. Since it is a good representation of how I am going to be writing the entire test plan, I thought I might run it by you to make sure it looks okay prior to diving into all the rest of the work:

https://github.com/lkcampbell/ifdb/blob/issue-011-fix/test/test-plan.md

Let me know what you think so far.

PR #80, opened 2021-03-17
Selenium Tests for No Login top bar and footer controls

lkcampbell:
This is the first part of addressing issue #11. It contains a Selenium IDE test for the top control bar navigation and the footer navigation with no login. I also added ids to the elements being tested since this is the easiest way to locate them during the automated tests. More to come as I continue adding to the test suites.


dfabulich:
Can we add documentation explaining how to run the tests? (I'd like to run them once before I merge.)

dougo:
IMO it's a bad idea to refer to id or class attributes in UI tests. My rule of thumb is that you should always be able to refactor the page implementation without breaking the tests as long as the visual appearance and behavior of the page stay the same.

lkcampbell:
> Can we add documentation explaining how to run the tests? (I'd like to run them once before I merge.)

Yes, I can do that today.

lkcampbell:
> IMO it's a bad idea to refer to id or class attributes in UI tests. My rule of thumb is that you should always be able to refactor the page implementation without breaking the tests as long as the visual appearance and behavior of the page stay the same.

Okay, what do you suggest to use as locators?

lkcampbell:
@dfabulich: Instructions have been added.

dfabulich:
Locators are a hard problem. We want the tests to keep working even if the page "changes," but it's hard to predict how the page will change.

On the one hand, maybe we'll want to redesign the top bar; after we're done, there could be a "menu" instead of a top bar. So perhaps it would be better to use a `text=Log In` locator.

On the other hand, maybe we want to change "Log In" to "Sign In." By using ID locators, we can leave the old ID as `topbar-login` and the tests will continue to work. (But then the locator's name would be confusing.)

My general feeling is that it's impossible-ish to make good locators that survive major refactorings, so I'm happy to merge what we have. If we ever do try to redesign the top bar, we'll probably have to rewrite these tests, but that's probably true regardless.

dougo:
@dfabulich: I think it's best to write the tests from the point of view of the user. The user doesn't care what the ID of the element is, so that shouldn't be part of the testing contract. But the user does care that the link says "Log In". If you accidentally erase or change the link text, but the ID stays the same, the test should fail, so that you know that you broke something that was part of the user expectation. But if the link still says "Log In", it should keep passing even if the ID changes or is removed, otherwise you just have more busywork to fix the test even though it should have been a behavior-preserving refactor.

dougo:
And I agree that it's sometimes hard to write a locator, which is why I said it's a rule of thumb. Just thought it was worth stating up front so we don't go too far down the road of using IDs everywhere. Or, feel free to outvote me, if you prefer the style of using IDs everywhere, though in either case I think it would be useful to have an intentional guideline stating that.

lkcampbell:
@dougo, following your line of thinking, pull request https://github.com/iftechfoundation/ifdb/pull/85 is a good, concrete example of where the `id` locator approach starts to fail.

As @dfabulich and you have been discussing, it is difficult right now to know what to test when we don't even completely know what IFDB is going to become. So, I am following the philosophies or compromise and baby steps for now.

One way I am doing this is to force myself to strictly use Selenium IDE for my local testing even though it pains me to not use Mocha, so that we at least have a basic testing structure that is easy for beginners to use regardless of local dev environment.

Selenium IDE is far from ideal for testing. I'm thinking of it as a baby step above a testing document that requires manual testing. And if -- more likely when -- we have to throw it out, it will at least provide the design structure for our new approach to testing.

The `id` locator approach also forces me to start writing `id` attributes into the UI, which could be helpful down the road if, or when, we make IFDB more dynamic. So, if nothing else, I believe the is a good thing.

As with every new project, I appreciate any team guidance and ideas because I don't know exactly what I am doing. Then again, that's never stopped me before :).

andrewschultz:
> As with every new project, I appreciate any team guidance and ideas because I don't know exactly what I am doing. Then again, that's never stopped me before :).

Hey, keep it up. When I see someone else taking on something new where they're not 100% sure, it pushes me to look into new stuff. I'd always put off learning about Selenium, and your jumping into things may convince me to. Heck, maybe I might be able to fill in some detail-testing down the road if you want to tackle bigger issues.

I know when learning something new, it's good to have projects where I already know a lot. I've certainly browsed/lurked on IFDB a lot over the years & so seeing your commits and documentation helps me to understand what's going on.

PR #82, opened 2021-03-17
Update "Baf's Guide ID" links in Game Details section

alice-blue:
Resolves https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/204

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1816,7 +1816,7 @@ if ($embargoCnt > 0)
                         title=\"Go to Baf's Guide to the IF Archive\">
                         Baf's Guide</a> ID:
                      <a class=silent
-                      href=\"http://www.wurb.com/if/game/$bafsid\"
+                      href=\"https://web.archive.org/web/20110100000000/http://www.wurb.com/if/game/$bafsid\"
                         title=\"Go to the Baf's Guide listing for this game\">
                         $bafsid</a>
                      <br>" ?>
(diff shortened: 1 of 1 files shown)


dfabulich:
Looks good to me!

I think you mentioned that you didn't know exactly how this worked, so, FYI, `helpWinHRef` is a PHP function, defined in [pagetpl.php](https://github.com/iftechfoundation/ifdb/blob/main/www/pagetpl.php#L59-L63) which is included at the top of every page on the site.

```php
function helpWinHRef($href)
{
    return "href=\"$href\" target=\"IFDBHelp\" "
        . "onclick=\"javascript:helpWin('$href');return false;\"";
}
```

Thus `<a class=silent " . helpWinHRef("help-bafs") . " title=\"What's a Baf's Guide ID?\">` renders like this:

```html
<a class=silent href="help-bafs" onclick="javascript:helpWin('help-bafs');return false;" title="What's a Baf's Guide ID?">
```

That `onclick` handler runs the `helpWin` function in JavaScript on the user's browser; it's defined in [ifdbutil.js](https://github.com/iftechfoundation/ifdb/blob/main/www/ifdbutil.js#L3-L7).

```js
function helpWin(url)
{
    win = window.open(url, "IFDBHelp",
                      'width=400,height=400,left=10,top=10,scrollbars=1,resizable=1');
}
```

Thus, this link will pop up an a mini window pointing to the `help-bafs` page when you click on it.

As for how I figured that out: I searched the code base for `function helpWinRef` and read it, then looked at the "view source" output of the page. Then, when I was curious to know what `helpWin` did, I searched for `function helpWin` and read that.

PR #83, opened 2021-03-17
Move "Tips & More Info" link from homepage to top menu bar

Iniquit:
Hi! It's exciting to see IFDB go open source! This PR addresses [ifdb-suggestion-tracker issue #116](https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/116). Also standardizes spaces between "curarrow" icon and the menu items next to it (made them all use a single space, like 'Home' uses now.)

**Before:**
before

**After:**
after

### DIFF
--- a/www/home
+++ b/www/home
@@ -65,7 +65,6 @@ pageHeader("The Interactive Fiction Database - IF and Text Adventures");
          recommendation engine.  IFDB is a Wiki-style
          community project: members can add new game listings, write
          reviews, exchange game recommendations, and more.
-         <span class=details><a href="tips">Tips &amp; More Info</a></span>
 
          <?php
 // -------------------------- inbox ----------------------------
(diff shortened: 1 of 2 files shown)


www/home
```diff
@@ -65,7 +65,6 @@ pageHeader("The Interactive Fiction Database - IF and Text Adventures");
          recommendation engine.  IFDB is a Wiki-style
          community project: members can add new game listings, write
          reviews, exchange game recommendations, and more.
-         <span class=details><a href="tips">Tips &amp; More Info</a></span>
```
dfabulich, line 68: I wonder if we should add this back? It's OK to have multiple links, especially if our goal is to make that page more visible.

dfabulich:
This commit is currently in conflict with the main branch _and_ with your responsive home page branch! 😬  We may want to let the responsive home page branch go through first, and then reapply this PR again later when the header is refactored and the code is settled down…?

www/home
Iniquit: Works for me!

Iniquit:
Sounds like a good idea! This one is super small anyway.

dfabulich:
I'm going to close this old conflicting PR. Feel free to reopen in the future.

PR #85, opened 2021-03-18
Responsive homepage

Iniquit:
This PR makes an initial pass at a number of mobile display bugs by converting the IFDB homepage and topbar to a responsive layout, using flexbox instead of tables.

(If the whole site can be made responsive, it'll no longer be necessary to maintain a separate mobile stylesheet-- the layout of pages should just adjust automatically. The custom mobile stylesheet could still be used afterwards we don't want to display some content on mobile.)

This PR also uses includes to break several pieces of display logic out of the main home.php file into a subdirectory provisionally labelled "components"-- this reduces the length of home.php from ~1200 lines to ~100 and should hopefully make it much easier to work on the homepage in the future.

This PR is based off the 'dev' branch to avoid re-implementing the changes that have already been introduced there.


Iniquit:
Rebased onto new dev branch.

dfabulich:
PRs should normally target the `main` branch. (My apologies; we don't have a `CONTRIBUTING.md` so there's no way you would have known that.)

The `dev` branch is the branch representing the code that's currently live on our internal "staging" server.

dfabulich:
This rebase looks like it failed; there's two copies your three commits. I tried to re-rebase it myself and pushed a `Iniquit-responsive-header` for you. Take a look at that and make sure I didn't miss anything? (And I think the new Selenium tests don't work in this branch either?) @lkcampbell 

I also think I want to request that we split even these commits up a bit further.

1. I certainly see why you refactored the home page into smaller pieces, but in a big refactor like this, we want the refactoring commit to _just_ split the home page up, with no further changes. Any commit that takes the home page and splits it up into pieces will be unreadably large, so it's preferable to make no additional content changes at the same time.
2. In addition, it looks like you did some streamlining of the CSS, e.g. removing the `div` tag name from a bunch of selectors. That's a good idea! But you're doing it in the same commit where you're refactoring the home page _and_ adding a flexer. This further obscures the "real" changes you're making in a sea of refactoring. So much is changing here that it's hard to see what's changing due to moving the code around and what's actual work that you're doing. (e.g. it looks like `.aside-box` moved around or something… but it's not clear why, or what that has to do with the rest of the commit)
3. Similarly, when pulling in the legacy mobile stylesheet, you're _both_ pulling it in _and_ reducing its scope; I think it'd be way better if those were two separate commits, perhaps _more_ than two. If I'm not mistaken, that commit alone is intended to fix half a dozen separate bugs by deleting bad rules, but it's hard to see what's going on there and why. If you really are fixing half a dozen bugs by removing half a dozen rules, I'd like each one of those removals to have a separate commit (with a separate commit message!) so I can understand your thought process for each fix.

OK OK OK, enough about commits. What about the code? I love that you're adopting a flexer here, and I think when we're done, the home page will be in a _much_ better place as a result. A couple of points:

4. Possibly due to my haphazard rebase, `topctl` no longer has any horizontal margin, so the "n" in "Log In" is glued right up onto the side of the bar. I think we want `topctl nav` to have `margin: 0 1em` or something.
5. The main branch put a little extra distance between the "Go" search button and the "Log In" button, presumably to prevent anyone from accidentally logging out or something? Seems like we ought to bring that back.
6. There doesn't seem to be any margin in the `.aside-box` boxes.
7. There's no margin or padding between the two `div`s in `topctl`, so if you make your browser kinda narrow, the centermost links are right up next to one another, like this: "Your InboxBrowse"
8. The "New to IF?" looks too narrow to me. You've got it at `15rem` right now (I generally prefer `ch` for horizontal sizing, though there's no "root" `ch` unit); I'd go with something more like `19rem` or `30ch`.
9. In mobile sizes, `topctl` wraps its links, making it difficult to click on "Browse" without clicking on "Home" accidentally.
10. I do worry that the `control-panel` box will be too hard to discover on mobile. It's after "New on IFDB" and after "Vote!" so it would be natural to assume "oh, looks like I've seen all there is to see; the rest of the page must be more of the same." I've mused that we may need a hamburger menu or at least some hierarchical approach to discovering these features.

Overall, I'm really impressed by how much work you've put into this (and I'm quite nervous about asking you to do more of it 😅) but I'm optimistic that we can work together to drag this page kicking and screaming into the 2020s!

dfabulich:
I put some more thought into this tonight and I've changed my mind… your placement of the "control panel" links is great, i.e. wayyyyy down at the bottom where nobody ever needs to see/find them!

The control panel is divided into four sections: Find, Browse, Contribute, and Personalize.
* Find: If you want to search for a game, just do it from the header. If you want to search for a list, poll, competition, club, or member, you can click the "Search" link in the header, which takes you to the "Advanced Search" documentation, and includes links at the top to search for different types of things. (Those links don't layout correctly on mobile, but we can fix that.)
    * The exception here is "Create a Poll," which, I agree is _kind_ of like "finding" a game, but not really. We should link to that page from the "Vote!" section https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/226 and when you're browsing/searching for a poll https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/227
* Browse: There's a "Browse" link in the header that takes you to the "Browse Games", with links at the top to browse by the other things. If the "Random" link moved out into the "IFDB Recommends" section, we wouldn't need the "Browse" section at all on the mobile home page.
* Contribute:
    * Review a Game: This takes you to a page where you can search for a game to review. (It also suggests a list of games that you've indicated that you've played but not yet reviewed.) This link doesn't need to be on the mobile home page.
    * Edit a Game Listing: This just takes you to a search page! I'm kinda inclined to delete this link altogether.
    * The rest of the links are "Create" links, which I think would make sense on their respective "Browse" pages, e.g. there's a link to create a game when you're browsing by game, a link to create a poll when you're browsing by poll, a link to create a list when you're browsing by list, etc.
* Personalize:
    * Edit your profile: There's a link to do this in the header.
    * Pick a custom display style: I think maybe this feature should go away entirely https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/199 but at the very least it doesn't need to be on the home page. (You can set these preferences by editing your profile)

In no case would I intentionally delete all of these links from the desktop home page, but it's now clear to me that I don't need to worry about people successfully discovering this section; instead, the links should be spread out into other, more discoverable places, and the control panel itself can park itself down at the bottom of the mobile home page, where it can think to itself about what it's good for.

Iniquit:
Whoops! I tried to get fancy and do the "rebase, then force-push" thing (but I think maybe if you pull the target branch down in the middle of that, it duplicates the commits?) I suspect just merging the upstream branch into my PR whenever it changes is probably easier and less error-prone-- is it okay if I do that in the future? Not 100% sure what the best practice is there.

Splitting up the commits is a good idea (and I probably should have been doing it already xD)! I'll definitely do that. Probably it'll be easiest for me to just start a new branch, step through it and replay my changes with better individual commit messages-- that should also make it compatible with the Selenium tests and anything else introduced in the meantime.

The points you mention on spacing and margin all make a lot of sense to me, and I'll work on correcting them in this second pass! Getting some space in there in the mobile header should prevent a lot of frustrating misclicks.

Awesome. Thanks a bunch for your patience and your guidance on this! I'm still pretty new to PHP and open-source in general, so I hope I can improve "on-the-job" while helping improve the IFDB!

dfabulich:
IMO, the "best practice" is to do an interactive rebase onto the main branch, and carefully write and rewrite your commits to "tell the story" about how the feature came into existence, in a way that makes life easier for reviewers.

Having said that, I'm not so picky as to insist on the very Bestest Practices™. If you just merge up from main periodically, and as long as each commit looks good on its own, I'll probably just merge your code as-is. If I feel lucky, I might dare to rebase your branch myself to strip out the merge commits, but probably not. (In this case, some of the commits were in an invalid state, so we wouldn't want to merge it like that.)

Iniquit:
Gotcha, makes sense! I'll aim for the interactive rebase then. Thank you!

lkcampbell:
> (And I think the new Selenium tests don't work in this branch either?)

@dfabulich, to clarify, is there a git/github problem with the Selenium IDE tests? Or do you mean that once this commit is merged, I need to update the tests. I'm assuming the later but there is a lot going on in this discussion so I want to make sure I understand.

dfabulich:
For the record, I'm attaching screenshots here of what my rebased version of the `Iniquit-responsive-header` branch looks like. (And only just now did I realize that I should have called the branch `Iniquit-responsive-homepage` … oops)

Click on the images for their full-size version

Iniquit:
@dfabulich Hi! I think this should be ready for review again. I tried to catch all the issues from earlier, but please let me know if I missed anything!

dfabulich:
I just warmed up the branch on my machine and it looks solid. I'm going to review the commits one by one (this might take me a bit) and get back to you. Thanks so much for all your hard work!

Iniquit:
It's my pleasure; thank you too!

dfabulich:
One oddity I just noticed is that the font size changes when I enable dark mode; the header font gets smaller. (In Chrome Dev Tools, you can test this with the "Emulate CSS prefers-color-scheme: dark" feature.)

Iniquit:
Oh, funny-- it looks like the dark-mode css actually has a different font-size. I think maybe it's cascading down and making everything a little smaller?
Regular CSS
```
body {
    background: #ffffff;
    font:  100%/140% Verdana, Arial, Helvetica, Sans-Serif;
    color: #000000;
    line-height: 1.4;
    margin: 0px 0px 1.5em 0px;
}
```
vs dark mode:
```
    body {
        background: #000000;
        color: #ffffff;
        font: 11pt/15pt Verdana, Arial, Helvetica, Sans-Serif;
    }
```

dfabulich:
I've got other work I need to do this afternoon, so feel free to toss in anything you'd like to to do today.

The dark mode spreadsheet was copied and pasted from MJR's "Midnight" custom style sheet; I can see no earthly reason why it should have any font size declarations at all.

Iniquit:
Sounds good! I'll whip up some fixes.

dfabulich:
You've done a heroic job here! Thank you so much. I have a few trivial comments to make, but I'm going to merge this commit now, regardless.

1. This branch marks a number of the main `<div>` sections of the home page with `class="block"` but there's no corresponding `.block` class definition in the CSS. The `class` seems to be doing nothing. Was it supposed to do something?
2. In a couple of places, the code uses `flex-grow: 999`, which is IMO a "code smell".
    * In `.main-nav ul`, I believe `flex-grow: 1` is sufficient here.
    * In `.col-main`, it's configured to `flex-grow: 999` to push against `flex-grow: 1` in `.col-sidebar`. But we don't want the sidebar to grow at all in desktop mode; we should set `.col-sidebar` to `flex-grow: 0` in desktop mode, and set it to `flex-grow: 1` in `legacy-mobile-styles.css`.
3. I think it would be appropriate to merge the `legacy-mobile-styles.css` into the main stylesheet now. (I think most user themes are around coloring/styling anyway.)
4. In a few places, you're using `0em`; my perspective on preferred style is to write `0` without units in that case.
5. I'll get over it, but in a few places this code cuddles the `{` up next to the selector, e.g. `.main-nav{` where I would prefer a space there.

Iniquit:
I'm glad I could help! Thank you too!

The `block` class was just something I was using to keep track of groups of elements inside flexers while reformatting the homepage-- it didn't have any semantic meaning and should probably be dropped unless that distinction ever becomes important.

The CSS points you raise make sense to me. I'll keep an eye out for those in future contributions.

I know some JS projects use eslint configuration files for standardizing code style and formatting-- I'm not sure if there's a PHP/HTML/CSS version, but I wonder if something like that could be useful for the project? Or maybe just adding code style guidelines to contributing.md to keep contributors on the same page.

lkcampbell:
@Iniquit thanks so much for your hard work on this. The refactored code and the adaptive functionality are both great! Now that this is working for the rather complex home page, using your ideas for the other pages will be much easier as well.

One point did come up in the IFDB advisor discussion regarding the UI for the top control bar. The original control bar looked like this:
e3bec08f3bb7c313c35a6ea66d44c70acf484826_2_1035x84

And now it looks like this:
9562c33041b653e963c12c69927523531c241691_2_1035x81

Glad you got replaced the arrow graphic with the bolded font, it looks cleaner.

Looks like we might have had some changes in the hyphens and vertical bars, however, which we are using to indicate the website structure. I'm pretty sure this wasn't intentional but I wanted to verify with you.

Again, fantastic work, thanks so much for your time and effort on this IFDB improvement.

Iniquit:
@lkcampbell Hi! You're welcome; I'm glad I could contribute! 

The removal of hyphens and vertical bars was intentional; the new styling on the topbar adds more spacing between the links (and also highlights their backgrounds on hover, although that won't be visible on mobile), so I figured using separators and underlining to indicate the divisions between the links wouldn't be necessary any more. I kept the two between Profile/Edit and Browse/Search because they were related to indicate some kind of connection ("Edit" doesn't make sense by itself) 

From what I can tell, a lot of modern websites take this approach in their navigation, with no text separators, trusting that the spacing will be enough to clue people in. On the other hand, if you think this would be confusing to IFDB's established userbase or just don't like it, please feel free to revert or alter it!

lkcampbell:
@Iniquit haven't received any feedback from others in the group regarding the hyphens and vertical bars in the top control bar, so I am going to assume it is okay as is.

PR #88, opened 2021-03-19
Allow admins to see helpful/unhelpful review votes

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/48

When logged in as an admin, each review has a link:

image

You can click that link to go to `/adminops?reviewvotes?reviewid=#####` to see who voted helpful/unhelpful for a given review. If you click on the user's name, you'll go to `/adminops?reviewvotes&userid=#####` to see a summary of that user's helpful/unhelpful votes, followed by a complete list of all of the reviews on which they voted helpful unhelpful.

You'll find that there's a bit of an issue with _testing_ this PR in local dev: the IFDB IFArchive scrubber scrubs the identities of all votes! Therefore, all reviews in local dev will say, "This review has no votes: 59797" even though it clearly _does_ have votes.

So, uh, all testing will have to happen on dev.ifdb.org.

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -127,6 +127,148 @@ if (isset($_REQUEST['burygame'])) {
     exit();
 }
 
+if (isset($_REQUEST['reviewvotes'])) {
+    if (isset($_REQUEST['reviewid'])) {
+        $reviewid = $_REQUEST['reviewid'];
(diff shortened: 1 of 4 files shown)


Durafen (approved):
Looking good.
Tested it with new votes.

Is there a reason for not making it public for everyone (and not only for admins)?

PR #90, opened 2021-03-20
Hidpi coverart

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/83

This is mostly about replacing direct references to game coverart with calls to my new `coverArtThumbnail()` function, which uses `` to show an image at an appropriate upscaled resolution. https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#resolution_switching_same_size_different_resolutions


Durafen (approved):
Looking good.

PR #96, opened 2021-03-21
Sort by Highest Rated on searches with no search terms

dfabulich:
For example, "published:2020" should sort by Highest Rated, because there are no search terms (no words) for relevance.

Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/19


Durafen (approved):
Nice, working as intended.

PR #97, opened 2021-03-21
Add HTML doctype

curiousdannii:


dfabulich:
This patch absolutely has an effect on layout; it's mostly pretty subtle, but in some spots it's quite noticeable, in an undesirable way.

https://developer.mozilla.org/en-US/docs/Mozilla/Mozilla_quirks_mode_behavior lists the effects of "quirks mode" (which is what we're using when we're not using a doctype).

Home page
1. topctl computed font size is 11.552px in this branch, 12.8px in main (I think this is because of the quirk "Don't inherit font properties into tables except for font-family.")
2. In the three `.aside-box` boxes (Reviewer Trophy Room, Join IFDB!, In the Database) the `<h3>` element has a `1em` top margin; in main, the margin collapses `<table>`, but in this branch, the margin appears inside the cell, creating a visible (and undesired) top margin. (I think this is because of the quirk "Collapse top margin of BODY and TD and bottom margin of TD.")
3. Something similar is happening with the `<p>` at the bottom of "IFDB Recommends." In main, its bottom margin collapses, but in this branch, the margin appears inside the cell, making a larger-than-necessary bottom margin

Game details

4. The font in the download box is 12.16px in main, 9.728px in the branch.
5. The `<h1>` top margin escapes/collapses in main, but is visible inside the `<td>` on the branch
6. Almost every individual line seems more spacious. (I think this is due to the quirk "Line height (not line-height) calculations are different" explained in the [almost standards mode](https://developer.mozilla.org/en-US/docs/Mozilla/Gecko_Almost_Standards_Mode) documentation. … but I'm not sure, it could just be the `<td>` margin thing, because IFDB uses tables _everywhere_.)

None of this is catastrophic, but I'd want somebody to go through and fix most of these issues before I'd want to merge this. (Maybe not #6? That one's  fine, I guess.)

Note also that @Iniquit is working on [refactoring the home page](https://github.com/iftechfoundation/ifdb/pull/85) and that would probably be impacted by this as well. I doubt it makes sense to work on both of these branches simultaneously; they're gonna conflict like crazy.

dfabulich:
I'm going to close this PR for now, but anyone who wants to work on these issues should feel free to file a new PR or request to reopen this one.

Issue #48, opened 2021-03-08
Remove IF Archive Mirror Preferences


lkcampbell:
## Repro steps:
1. Log into IFDB as a regular user.
1. Edit your profile.
1. Go down to section `5. Download Preferences`.
1. Open the `IF Archive Mirror` dropdown menu, select `ifarchive.smallwhitehouse.org (USA)` and `Save Changes`.
1. View the game [Anchorhead](https://ifdb.org/viewgame?id=op0uw1gn1tjqmjt7) and try to download `AnchorheadDemo.gblorb`.

## Observed behavior:
The download fails because it is trying to download from `ifarchive.smallwhitehouse.org` which doesn't exist any more as far as I can tell.

## Expected behavior:
There should not be a mirror dropdown menu. All traffic should be sent to `ifarchive.org`.

## Note:
This doesn't currently affect many of the download links because the functionality to change the links based on the `IF Archive Mirror` profile setting isn't functioning properly. See issue https://github.com/iftechfoundation/ifdb/issues/18 and https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/1 for more information. That's why I had to specifically pick the demo link for Anchorhead to demonstrate the problem.

Durafen:
Mirrors are saved in `mirrors` sql table, delete the row there will eliminate it.

erkyrath:
The whole mirrors system is out of date. `mirror.ifarchive.org` is no longer a random redirector; it's exactly the same as `www.ifarchive.org`. They're both handled by Cloudflare.

erkyrath:
I guess the action here is "remove the mirrors dropdown and send all traffic to `ifarchive.org`."

lkcampbell:
> I guess the action here is "remove the mirrors dropdown and send all traffic to `ifarchive.org`."

Edited Expected Behavior above to reflect this suggestion.

lkcampbell:
Going to wait until https://github.com/iftechfoundation/ifdb/pull/67/files goes in because some of the mirror code ties into plugin code that is soon to be gone.

lkcampbell:
Fixed by pull request https://github.com/iftechfoundation/ifdb/pull/99. Closing.

PR #99, opened 2021-03-21
Get rid of IFArchive mirrors

### LINKED ISSUE #18: http://ifarchive.org links aren't being converted to HTTPS
https://github.com/iftechfoundation/ifdb/blob/main/www/mirrorUrl.php#L73

We have code to convert download links from `http://www.ifarchive.org/if-archive/` into `https://mirror.ifarchive.org/if-archive/` but it only works with `www.ifarchive.org`. If the link starts with `http://ifarchive.org/if-archive/` without the `www.` it doesn't get converted.

That's bad, because Chrome blocks non-HTTPS download links by default.

dfabulich:
IFDB allows you to choose an IFArchive mirror, based on a short list of mirrors stored in the `mirrors` table; all three of those mirrors are wrong:

```
http://www.ifarchive.org/if-archive/
http://ifarchive.smallwhitehouse.org/if-archive/
http://mirror.ifarchive.org/if-archive/
```

`smallwhitehouse.org` has been down for a long time. The other URLs kinda work, but they should be using the canonical HTTPS URL, `https://ifarchive.org/`

Searching through the DB, these are the links I found. All of them are correctly converted:

```
http://www.ifarchive.org/if-archive/
http://mirror.ifarchive.org/if-archive/
http://ifarchive.org/if-archive/
http://www.ifarchive.org/indexes/

http://ifarchive.smallwhitehouse.org/if-archive/
http://ifarchive.jmac.org/if-archive/
http://ifarchive.ifreviews.org/if-archive/

https://www.ifarchive.org/if-archive/

https://ifarchive.org/indexes/if-archive/
https://ifarchive.org/if-archive/

https://www.ifarchive.org/indexes/
http://ifarchive.org/indexes/
```

This PR fixes #18.


lkcampbell:
Looks good to me. My approach was to drop the `mirrors` database as well and all refers to it but this is a less invasive approach.

Durafen (approved):
Looking good.

lkcampbell:
This pull request fixes issue https://github.com/iftechfoundation/ifdb/issues/48 as well.

PR #100, opened 2021-03-21
Wrap Baf's guide reviews in <p> tags so they wrap properly

dfabulich:
This limits the width of Baf's guide reviews to `60ch`.

https://ifdb.org/viewgame?id=op0uw1gn1tjqmjt7

image


Durafen (approved):
Looking good.

PR #101, opened 2021-03-22
Make the header stick to the top of the screen

dfabulich:
I'm… not sure if I like this. It seems especially wasteful on iOS, where there's a one-tap shortcut to scroll up to the top of the screen, and where the header bar is currently particularly useless, due to https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/216 and https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/217.

(Click for full-size images)

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -150,6 +150,12 @@ h3 {
     margin-bottom: 0px;
 }
 
+header {
+    position: -webkit-sticky;
+    position: sticky;
+    top: 0;
+}
+
 /*
 The "topbar" division is the area where we display the main IFDB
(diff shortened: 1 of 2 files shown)


dfabulich:
@alice-blue I've invited you to join this Github org and review the PR. The code doesn't strictly need review, but I'm curious for you to pull down the branch on your machine and see if you like it.

dfabulich:
My heart says that this PR is probably not good enough to merge as-is. I'm open to considering other PRs, but I'm going to close this one until we have another better idea.

PR #104, opened 2021-03-25
Hide private emails in the adminops user list

alice-blue:
Related to https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/46 and https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/265

In the user list on the adminops page, hide user's private email addresses behind a spoiler button. (The background color of the spoiler button doesn't exactly go with this table, with its alternating color rows, but I don't know what to do about that.)

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -770,6 +770,8 @@ else if (isset($_REQUEST['users'])) {
              $created, $login) =  mysql_fetch_row($result);
         $name = htmlspecialcharx($name);
         $email = htmlspecialcharx($email);
+		$email = spoilerWarning($email, "Click to see email");
+		$email .=spoilerWarningScript();
         $pubmail = htmlspecialcharx($pubmail);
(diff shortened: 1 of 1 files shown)


dfabulich:
I think this is a good start, but the email address is still showing in so many places, it might only be a first step. e.g. if you search for users, the email is showing there http://localhost:8080/adminops?finduser=rick&go=Find and of course if you click on any user, the email is showing there, too.

By contrast, for example, Discourse hides the email addresses in all of those places; it requires a click to see them.

Should I just merge this as-is? Or should we do more work first?

dfabulich:
There was a tiny bit of inconclusive discussion about this with the moderators. I currently think that this is basically the right thing, and so I'll merge it tomorrow-ish unless I hear outcry.

I do mildly prefer a version that allows you to click one button to “Show all emails” in the report, especially if you are searching the user list by a partial email address or something. But that could be done in a separate PR, if ever.

PR #113, opened 2021-03-29
Force browsers to revalidate CSS and JS cache

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/251

This is deployed to dev.ifdb.org. If you view https://dev.ifdb.org/ifdb.css in Chrome dev tools you can see it now has a `Cache-Control: no-cache` header.


lkcampbell (approved):
Confirmed the cache-control header is set correctly to "no-cache". Changes look good. Thanks.

PR #114, opened 2021-03-29
Add list of links to side panel on home page

alice-blue:
Add two boxes in the right panel of the home page: one box with links to IF websites, and one box with links to competition websites.

Not sure of the best way to specify margins/padding.

Closes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/173

### DIFF
--- a/www/components/database-stats.php
+++ b/www/components/database-stats.php
@@ -1,10 +1,11 @@
 <div class="aside-box">
     <h3>In the Database</h3>
 
+    <ul>
     <?php
     $result = mysql_query("select count(*) as c from games", $db);
     $cnt = mysql_result($result, 0, "c");
-    echo "&bull; <a class=silent href=\"search?browse\">$cnt Game Listings</a><br>";
+    echo "<li><a class=silent href=\"search?browse\">$cnt Game Listings</a></li>";
 
     $result = mysql_query(
         "select count(*) as c from reviews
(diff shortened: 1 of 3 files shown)


dfabulich:
Ah, _now_ I see what you were wrestling with here. I would have expected this box not to need to adjust margin/padding at all, but if we don't, the bullets look super weird when juxtaposed with the "In the Database" box.

image

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Consistent_list_indentation is a fun article about list indentation. It's _weird_. The bullet is designed to appear outside the `<ul>`s box, and so, to make it visible, browsers (not just Gecko—it's Chrome, too) add in `40px` of padding on `<ul>` elements.

But that's actually a really weird amount of space, because the bullet's spacing from the text is defined not in `px` but in `ch` units. So if you crank your font size up to a huge size, like this, the `40px` buffer is insufficient, and the bullet disappears off to the left, like so:

```html
<html>
<div style="font-size: 256px">
  <ul><li>bullet</li></ul>
</div>
</html>
```

IMO, the right way to make sense of this is to set the left padding of `<ul>` to some number of `ch`. `4ch` is pretty reasonable when the bullets appear juxtaposed with paragraphs `<p>`, but here I think it makes sense to use just `3ch`.

Let me take a crack at this.

dfabulich:
I made a pull request targeting your pull request branch. https://github.com/alice-blue/ifdb/pull/1 (This is weird, but I think it's the right thing in this case.) LMK if you have questions.

alice-blue:
OK, I merged it. Thanks.

PR #116, opened 2021-03-30
Update editprofile page to say "Settings"

alice-blue:
Update editprofile page to say "Settings" in a few places instead of "Profile," to be consistent with the name of the link in the navigation bar.

I do not know what the "editprofile.email" part of the following code does, so I left that as is:
 `pageHeader("Update Settings", "editprofile.email", false,`

Related to https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/254

### DIFF
--- a/www/editprofile
+++ b/www/editprofile
@@ -619,14 +619,14 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST'
 }
 
 // start the page
-pageHeader("Update Profile", "editprofile.email", false,
+pageHeader("Update Settings", "editprofile.email", false,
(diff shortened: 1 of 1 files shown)


dfabulich:
The `pageHeader` function is defined in https://github.com/iftechfoundation/ifdb/blob/main/www/pagetpl.php#L71 where it says that the second parameter is `$focusCtl`. That gets passed to `basePageHeader`, which then adds a bit of `onload` JavaScript to focus the element defined in `$focusCtl`. In this case, the email address field is focused by default, and that's just fine.

PR #120, opened 2021-03-30
Fix markup errors in game page

salty-horse:
* Fixed unclosed tags.
* Added missing space between attributes.


dfabulich:
How are you finding these errors? Are you using an HTML validator or something? (Which one?)

PR #128, opened 2021-04-01
Add ability to search for game tags

lkcampbell:
Fix for https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/35.

### DIFF
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -938,7 +938,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
     // run the query
     $result = mysql_query($sql, $db);
     if (!$result) error_log(mysql_error($db));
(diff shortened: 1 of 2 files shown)


www/searchutil.php
```diff
@@ -204,7 +204,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
     {
         // special keywords for tag search
         $specialMap = array(
-            "tag:" => array("tag", 99));        
+            "tag:" => array("tag", 0));
```
dfabulich, line 207: What's this about?

lkcampbell: Yeah, we should probably assign constant names to those magic numbers to make the code more readable.

I was using 99 for special-case handling, playing around with different ideas for the tag search. Missed this when I was cleaning up after myself. The 0 is for text matches.

PR #129, opened 2021-04-01
Fix glitch when hovering out of star ratings

salty-horse:
When moving the mouse from right to left over the "Rate It" control, until the cursor is no longer over the stars, it would be stuck in 1 star instead of reverting to the set rating.

Changed to use a single stars `` element, and trusting the `mouseout` event.

I also removed usage of deprecated javascript: `document.all` and `document.captureEvents`. I think it was a workaround for old browsers, and is not needed now. `document.all` isn't used anywhere else.

### DIFF
--- a/www/starctl.php
+++ b/www/starctl.php
@@ -53,40 +53,7 @@ function setStarCtlValue(id, val)
         ?>
 <script type="text/javascript">
 <!--
-var starRatings = [];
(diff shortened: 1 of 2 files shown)


dfabulich:
Interestingly, I was only able to reproduce this bug in Firefox, and not in Safari or Chrome.

PR #133, opened 2021-04-04
Link to intfiction.org on the Create a Poll page

alice-blue:
People sometimes use polls to ask questions that aren't a good fit for IFDB polls.

This adds a tip box to the Create a Poll page linking to the forum so people can ask questions there instead.

This uses the "tipbox" class, but the layout doesn't look great on small screens. Is there a better way?

Resolves https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/224

### DIFF
--- a/www/poll
+++ b/www/poll
@@ -838,6 +838,10 @@ if ($id == 'new') {
     pageHeader("Create a New Poll", "newPollForm.title");
 ?>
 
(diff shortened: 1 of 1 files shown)


dfabulich:
Well, tip boxes don't even display on mobile, which is the most typical small-screen use case. I think it looks fine when I just narrow my browser width.

PR #134, opened 2021-04-04
Adopt `box-sizing: border-box`

dfabulich:
Fixes iftechfoundation/ifdb-suggestion-tracker#247
Fixes iftechfoundation/ifdb-suggestion-tracker#248

<https://www.jefftk.com/p/the-revenge-of-the-ie-box-model> explains more about how `box-sizing: border-box` came to be.

<https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing> explains how to use it.

I'm using the recommended best practice documented here: <https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/>

That makes everything use `box-sizing: border-box` except for sections that we deliberately change to `box-sizing: content-box` as needed. (I clicked around and couldn't find any issues, but if we find them, they'll be a one-line fix.)

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -57,7 +57,13 @@ CSS unfortunately doesn't have a way to refer indirectly to the dimensions
 from the image file, so you just have to hard-code the actual pixel size.
 */
 
+html {
+    box-sizing: border-box;
+}
 
+*, *:before, *:after {
+  box-sizing: inherit;
+}
 
 /*
 This is the style for the main body of the page - nearly everything
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -929,7 +929,7 @@ p.nosp {
     margin: 0 0 0 0;
 }
 
-p {
+p, ul, ol {
     max-width: 60ch;
 }
 


dfabulich:
@alice-blue I've deployed this to the dev server; take a look.

<https://dev.ifdb.org/tos>
<https://dev.ifdb.org/code-of-conduct>

alice-blue:
Yeah, this makes those two pages look better.

alice-blue:
@dfabulich I'm noticing that dev.ifdb.org looks different from the live site, in terms of the banner width vs. the width of the rest of the page. On the live site, the widths match. On the dev site, the banner is a little wider than everything else. (Not sure if it's a result of this PR or not.)

Live site:
ifdb banner width live site

Dev site:
ifdb banner width dev site

(And I think this is not related to this PR, but....) I'm not actually sure how the text is supposed to be matching up, either (e.g. should "Home" in the navigation bar be aligned with "Welcome" below it?)

dfabulich:
Good catch! I'll investigate

PR #135, opened 2021-04-04
Consolidate mobile styles

dfabulich:
`legacy-mobile-styles.css` activates when IFDB detects a phone via user-agent sniffing, but it includes a section at the top that says `@media only screen and (max-device-width: 480px) {` which means that those styles could just be added to the main CSS rules without UA sniffing.

So, I've done that here, moving rules out of the legacy mobile stylesheet and into the main set of rules in as idiomatic a way as I can.

Along the way, this partially helps with https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/213 and fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/287

I recommend reviewing this commit-by-commit.

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -218,6 +218,7 @@ div.topctl form {
     flex-wrap: wrap;
     overflow: auto;
     padding: 0em 1em 0em 1em;
+    justify-content: space-between;
 }
 
 .main-nav ul{
@@ -225,7 +226,6 @@ div.topctl form {
     padding: 0px;
     display: flex;
     flex-basis: 0;
-    flex-grow: 999;
     align-items: center;
 }
 
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -741,7 +741,10 @@ div.edit-popup-win table tr td {
 div#imageUploadCopyrightForm table td {
     font-size: 90%;
 }
-        
+
+input#searchfor, #searchfor2{
+  max-width: calc(100vw - 2rem);
+}
 
 /*
 The "searchSummary" table is used to show the current location in the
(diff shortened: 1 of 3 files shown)


www/ifdb.css
```diff
@@ -774,13 +765,23 @@ div.searchHelpList td {
...
+    margin-left: -1rem;
+    padding-left: 1rem;
+    text-size-adjust: none;
+    -webkit-text-size-adjust: none;
+    flex-wrap: wrap;
 }
 div.browseTabs {
-    padding: 6px 0 0 4px;
+    padding: 6px 1em 0 4px;
+    overflow: scroll;
```
lkcampbell, line 783: Looks like this is causing a partial white frame in my Mac Chrome browser:

Screen Shot 2021-04-06 at 7 23 46 AM

lkcampbell: Everything else looks good. I tested in Chrome with the device emulator. Also, checked a couple of the improvements at the dev site on on my iPhone.

dfabulich:
Yeah, it's even worse on Windows. I think I'll not merge this PR this week, probably.

image

www/ifdb.css
dfabulich: I removed that bit in the latest version of the PR. https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/213 is no longer fixed in this PR.

lkcampbell:
@dfabulich login spacing looks good and half-frame around the browsing bar is gone.

Assuming that @Iniquit is going to be addressing the browse/search bar wrapping issues per the IFDB Tech discussion, I would say this pull request is ready to be merged.

PR #136, opened 2021-04-04
Rename "Your Inbox" -> "Inbox", "Your Activity" -> "My Activity"

dfabulich:
This just barely fits on an iPhone 12 Mini
simulator_screenshot_F3A0C51B-BCD7-4197-83E7-B3D402430190


dfabulich:
@alice-blue I think we could bring back "Your Activity" if we got rid of "Home" as a text link…? The banner is clickable now.

alice-blue:
"Your" vs. "my" isn't a big deal to me either way.

alice-blue:
@dfabulich Are you planning to keep it as "my" or is that something that might change (if you take off "home")? I ask because I'm wondering if the headings on the activity page should change from "your" to "my" to be consistent.

dfabulich:
I think the first thing to do is to make a hamburger menu. https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/252 Then, we'll have enough room to call it whatever we want.

PR #137, opened 2021-04-04
Sort lists, polls, and clubs by newest by default

dfabulich:
The home page links to browse lists, polls, and clubs all sort by newest with an URL parameter, but if you click on Lists, Polls, and Clubs on the Browse page, it uses the default sort orders, which currently sort by title, and should sort by newest instead.

### DIFF
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -811,20 +811,20 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
 
     case "list":
         $sortList = array(
-            'ttl' => array('title,', 'Sort by List Title'),
-            'usr' => array('username,', 'Sort by List Author'),
             'new' => array('moddate desc,', 'Newest First'),
             'old' => array('moddate,', 'Oldest First'),
+            'ttl' => array('title,', 'Sort by List Title'),
+            'usr' => array('username,', 'Sort by List Author'),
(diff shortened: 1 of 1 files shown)


www/searchutil.php
```diff
@@ -811,28 +811,28 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
 
     case "list":
         $sortList = array(
+            'new' => array('moddate desc,', 'Newest First'),
```
lkcampbell, line 814: In your other changes, Oldest First follows Newest First, but not in this change.

lkcampbell: I noticed the Competitions browse list uses "Most Recent First". Suggest changing it to "Newest First" to be consistent.

lkcampbell (changes requested):

lkcampbell:
@dfabulich, when I am doing reviews for you and the suggested changes are simple like these, do you prefer that I make the changes as part of the review, or just make the suggestions and leave the changes for you to make?

dfabulich:
Github has a nifty feature where you can "suggest changes" and edit the code in a comment; if I agree, I can then click a button to apply your suggestions. For small changes like renaming "Most Recent First" that's the ideal thing.

I'll just make the changes myself this time.

PR #141, opened 2021-04-06
Fix padding for box-sizing: border-box

dfabulich:
@alice-blue [reported this issue](https://github.com/iftechfoundation/ifdb/pull/134#issuecomment-813777382); this PR restores it to look like the live site.

Live site:
ifdb banner width live site

Dev site:
ifdb banner width dev site

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -260,7 +260,7 @@ margins here.
 */
 div.main {
     max-width: 100ch;
-    padding: 1em 1em 1em 1em;
+    padding: 1em 0;
     margin: 0 auto;
 }
 


dfabulich:
Actually, after playing with this a while, I think I prefer it the new dev site way.

1. We certainly want some padding/margin on the `main` element in mobile mode, and indeed the live main site already has that. You can see if you gradually expand the window that the margin vanishes just as the window gets wide enough.
2. So the question then becomes: at what point do we want the main margin to disappear, and my intuition is "never."

@alice-blue WDYT?

dfabulich:
I'm leaving this as it is, for now

PR #142, opened 2021-04-07
Updated favicon

Iniquit:
Here's the PR for the updated favicon to gauge how it looks in everyone's browsers. I'm using https://realfavicongenerator.net to try to catch all the different contexts in which the favicon might appear. Their tool automatically generates a number of extra versions to make the favicon play nice on mobile homescreens, new tab pages, etc. I don't _think_ those should cause issues, but something to keep in mind.


dfabulich:
Deployed to https://dev.ifdb.org/

www/pagetpl.php
```diff
@@ -23,8 +23,16 @@ function basePageHeader($title, $focusCtl, $extraOnLoad, $extraHead,
 ?>
 <html>
 <head>
-   <link rel="icon" type="image/png" href="/ifdb-favicon.png">
-   <link rel="shortcut icon" href="/favicon.ico">
+    <link rel="apple-touch-icon" sizes="180x180" href="/favicons/apple-touch-icon.png">
```
dfabulich, line 26: It's a bit odd to see this here… it's the icon used for "add to home screen" (a2hs) on iOS, supposedly, but it doesn't seem to work when I a2hs on my iPhone 11 running iOS 14.4.2.

www/favicons/site.webmanifest
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "",
```
dfabulich, line 2: This manifest is used for "add to home screen" (a2hs) on Android. If we're gonna have one, we should put a real `name` and `short_name`. I think "IFDB" would work great for both.

www/favicons/site.webmanifest
```diff
@@ -0,0 +1,19 @@
...
+        },
+        {
+            "src": "/favicons/android-chrome-512x512.png",
+            "sizes": "512x512",
+            "type": "image/png"
+        }
+    ],
+    "theme_color": "#ffffff",
+    "background_color": "#ffffff",
+    "display": "standalone"
```
dfabulich, line 18: `"display": "standalone"` means "don't show the URL bar." IMO that's an unwise choice for any site that isn't designed to handle it. For example, there would be no obvious way to use the back button if you a2hs. I think `"display": "minimal-ui"` or `"display": "browser"` would be more appropriate.

www/pagetpl.php
dfabulich: ohhh, I see the problem. Safari refuses to use basic auth when downloading a favicon.

www/favicons/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<browserconfig>
+    <msapplication>
+        <tile>
+            <square150x150logo src="/favicons/mstile-150x150.png"/>
+            <TileColor>#603cba</TileColor>
```
dfabulich, line 6: That seems like an odd choice of `TileColor` … lavender?

image

www/favicons/site.webmanifest
Iniquit: Sounds like a plan!

www/favicons/site.webmanifest
Iniquit: Got it, thanks! I confess I basically just followed their wizard and so didn't know what all of these stood for. Maybe just `"display": "browser"` for now since it appears to be straightforward and we can be reasonably certain it won't break anything.

www/favicons/browserconfig.xml
Iniquit: It was one of a couple of choices in the wizard, possibly based on some set of Windows defaults. Should we go with white for now?

dfabulich:
Instead of renaming the previous favicons, I think we should just delete them. They'll live on in git history for anyone who's curious to see them. (Github's code review feature doesn't seem to offer a way to directly comment on a rename.)

Iniquit:
Merged in @dfabulich's improvements!

dfabulich:
Merged by squashing

PR #145, opened 2021-04-09
Only show top five games in author profile

lkcampbell:
This is a fix for https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/289.

Note: The `See all games...` link shows up regardless of the total number of games the author has written. I did this intentionally to keep the UI consistent but it should be addressed in the future. The "See all..." links should only show up if there are more than five of any given thing: games, polls, lists, or review/ratings. Note that the `Play Lists` section does the "See all..." links correctly.

### DIFF
--- a/www/showuser
+++ b/www/showuser
@@ -471,8 +471,7 @@ function pingMultiUserSession(sid, gameid, title, author, started)
          where
            gameprofilelinks.userid = '$quid'
          group by games.id
-         order by starsort desc
-         limit 0, 6", $db);
+         order by starsort desc", $db);
 
     if (($cnt = mysql_num_rows($result)) > 0) {
         echo "<h2>Interactive Fiction by $username</h2>"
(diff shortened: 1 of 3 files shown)


lkcampbell:
Hold up on this, I found a bug.

lkcampbell:
Okay, fixed.

www/showuser
```diff
@@ -523,6 +524,9 @@ function pingMultiUserSession(sid, gameid, title, author, started)
                     . "<i>$desc</i></span></div>";
             }
         }
+        
+        echo "<p><span class=details><a href=\"search?authorid=$uid\">"
+            . "See all of this member's games</a></span>";
```
dfabulich: I think we only want to show this link if there are more games to see.

I also think it might be nice to show a count and the author's name in this link, e.g. "See all 7 games by Author Name"

lkcampbell: Okay, I was originally going to fix all of the "See all..." links on the page as a separate issue, but if you want me to push it into this pull request, I can do that.

I like the count idea as well.

lkcampbell:
I pushed the page fix to only show "See all..." link if there are more than five things in the given section. There are a couple of special cases.

First, the reviews list has two "See all..." links, one for reviews and one of reviews and ratings. I hid both links if there are less than five reviews only. If this is not the behavior you are expecting, let me know and I can change it.

Second, the clubs list. There is no "See all clubs" page in the site as far as I can tell. The list of clubs will simply be the entire list of clubs. I don't think this will be an issue since I seriously doubt any IFDB member has joined more than five clubs.

Next, I am working on adding the list count and the member's name to the "See all..." link text.

lkcampbell:
Everything is done, thanks.

PR #146, opened 2021-04-10
Don't attempt to read or write mirror ID when editing profile

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/291

The problem was that we were attempting to set a mirror ID to '' which is invalid. (The mirror ID is now obsolete.)

This was a regression from https://github.com/iftechfoundation/ifdb/pull/99

I wasn't able to reproduce this problem in local dev, but I was able to reproduce the problem on dev.ifdb.org. It's working now on dev.ifdb.org.

### DIFF
--- a/www/editprofile
+++ b/www/editprofile
@@ -154,7 +154,7 @@ function init()
            picture is not null as haspic, picture as pic,
            acctstatus, activationcode,
            concat(defaultos, '.', ifnull(defaultosvsn,'')) as defaultos,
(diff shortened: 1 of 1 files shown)


lkcampbell (approved):
The changes make sense to me. Looks like you hit every place where the database queries use mirrorid. I can't confirm it fixes the bug locally without a repro, but if you were able to repro the bug on dev and then make sure the bug went away after the fix, that's good enough for me.

PR #147, opened 2021-04-11
Don't personalize recommendations on game details page

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/223

It's not very clear to me why, but personalizing these recommendations makes the query take 15+ seconds in local dev, and milliseconds without personalization.

This might be a bit excessive; I don't think the queries for recommended lists and/or polls were very impacted by personalization. The main pain point was excluding ratings belonging to the current user when calculating recommended games, and excluding games that the current player had added to their playlist, wishlist, or unwishlist.

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1445,9 +1445,6 @@ if (!isEmpty($genre)) {
         // who rated this game highly.
 
         // look for lists containing this game
-        $sortMeLast = "";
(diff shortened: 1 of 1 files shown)


dfabulich:
nvm I found a cleverer way

PR #148, opened 2021-04-11
Improve recommendation performance on game details page

dfabulich:
The old query was doing a bunch of joins to exclude games from the current user's played games, wishlist, and unwishlist. This query grabs all of those game IDs with a subselect, which improves game details performance 100x in my local testing, from 15 seconds down to 0.15 seconds. (But I could never reproduce performance THAT bad on dev.ifdb.org)

(I don't even understand the purpose of joining to the reviews table a third time as long as we're excluding the current user from r1 and r2; I hope this doesn't introduce a weird regression.)

Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/223

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1514,26 +1514,16 @@ if (!isEmpty($genre)) {
         // play list or wish list - they obviously already know about
         // these games, so there's no reason to recommend them.
         if ($curuser && !$cssOverride) {
-            $joinOtherUser = "left outer join reviews as r3"
-                             . "  on r3.gameid = r2.gameid"
-                             . "  and r3.id != r2.id"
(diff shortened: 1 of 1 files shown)


lkcampbell (approved):
The performance is *so* much better. Much easier to read as well.

Yeah, I can't figure out what the r3 join is about either. I mean, it appears to be checking that the current user's ratings are not included as part of the recommendations, but the where clause makes sure r2.userid is not equal to the current user id, so maybe it was just redundant code.

Anyway, the changes look good to me and it's working great.

salty-horse:
@dfabulich can you delete the `details-recommendation-performance` branch if you're done with it?

PR #149, opened 2021-04-11
Allow more than one crossrec for each game

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/4

I've run this on dev.ifdb.org


Durafen (approved):
Looking good, this should do the trick.

lkcampbell (approved):
Yup, looks good and works well on my local machine. I'm excited to see this change because it is going to help improve recommendations quite a bit once people start using it more.

dfabulich:
I've now run this schema change against production ifdb.org

Issue #84, opened 2021-03-18
Suggestion: Add a regular test user to the local dev database


lkcampbell:
If we add a test regular user to the local dev database, it saves a step of registering and confirming a new user via the Docker logs.

Might make automated testing a bit easier as well once we get to that point. Of course, we still have to figure out how to test registering a user, but at least this is a step in the right direction.

lkcampbell:
Not sure the `enhancement` label is the correct one for this issue, but it was the closest I could find in the existing list.

PR #152, opened 2021-04-15
Add regular test user to the local dev database

lkcampbell:
This is a fix for issue https://github.com/iftechfoundation/ifdb/issues/84.

Adds a test regular user named `Test Tester` to the local dev database so you don't have to go through the extra steps of registering and the activating a test user.


dfabulich:
Thanks!

PR #153, opened 2021-05-07
Fix detection of HTML character references in fixDesc

salty-horse:
`fixDesc` currently generates a warning when mistaking the output of `preg_match` to be an array instead of string:

```
php > include_once "www/util.php";
php > echo fixDesc("<p>&#11354;</p>");
PHP Warning:  strlen() expects parameter 1 to be string, array given in www/util.php on line 1311
<p>&#11354;</p>
```

This commit fixes it, and also adds detection of hex-style character references such as `&#x1F354;`.
Note that both types can have any number of digits, but I only increased the current code's limit by one, which should be enough for emojis.

### DIFF
--- a/www/util.php
+++ b/www/util.php
@@ -1299,7 +1299,7 @@ function fixDesc($desc, $specials = 0)
         case '&':
             // if it's &lt;, &gt;, &#xxxxx;, &quot; or &amp;, leave it;
             // otherwise convert the & to &amp;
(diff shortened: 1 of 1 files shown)


dfabulich:
Sorry to keep you waiting so long! Tested with "Woman Vampire: Dark Skin Tone" 🧛🏿‍♀️ on a user profile and it worked great.

PR #154, opened 2021-05-08
Use consistent newline style in code

salty-horse:
Following discussion in iftechfoundation/ifdb-suggestion-tracker#305:
* Run dos2unix on text files to convert from CRLF to LF newlines.
* Make sure there's a newline at the end of text files.


NilsFagerburg:
As long as we're doing a big history polluting change, perhaps it would also be worthwhile to strip trailing whitespace and convert tabs to spaces.

salty-horse:
I will wait until all the other active PRs are resolved before rerunning the cleanup.

dfabulich:
Ok! There are no outstanding PRs except yours. Now is the time to land a big PR like this one.

salty-horse:
Done. I added another step that trims trailing whitespace from text files.

Diffing with `diff main --ignore-space-change` shows all of the changes only change whitepsace, and that no binary files were affected (I searched for mentions of `Binary`). I also made sure not to touch svg files.

PR #159, opened 2021-05-27
Strip parentheses in getSortingPersonalName()

NilsFagerburg:
This fixes bug iftechfoundation/ifdb-suggestion-tracker#277 and also
allows searches for "Real Name (as Pseudonym)" to return relevant
results.

### DIFF
--- a/www/util.php
+++ b/www/util.php
@@ -1959,6 +1959,12 @@ function splitPersonalNameList($name)
 //
 function getSortingPersonalName($name)
 {
+    // strip parentheses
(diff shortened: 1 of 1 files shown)


dfabulich:
Sorry to keep you waiting so long!

IMO, the code in searchutil.php to try to parse out people's last names looks is extremely fragile. This PR removes not just the parentheses, but also the contents of the parentheses, which is probably fine, since we're searching for a last name just by counting spaces… 🤷‍♂️

PR #160, opened 2021-05-27
Generate thumbnails with transparent background

NilsFagerburg:
Maybe fixes iftechfoundation/ifdb-suggestion-tracker#67
I was unable to test it completely (I can't save cover art in the test instance) but it seems to work (the generated tempimage thumbnail has a transparent background).


dfabulich:
Looks great on my test machine! Sorry for the long wait!

PR #161, opened 2021-05-28
Trim tags so they won't be pushed off screen by long download lists

NilsFagerburg:
Fixes iftechfoundation/ifdb-suggestion-tracker#309


dfabulich:
Thanks!

PR #162, opened 2021-05-28
Fix buggy mysql function calls

salty-horse:
These were found with the static analysis tool [Psalm](https://psalm.dev).

1. Several mysql functions were missing the crucial $db parameter.
2. An error-handler was called with an errant `$`.
3. `mysql_connect` was called with an extra (ignored) parameter.

I didn't test any of these.

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -108,7 +108,7 @@ if (isset($_REQUEST['burygame'])) {
                     set flags = '$newflags'
                     where id = '$qid'", $db);
                 if (!$result) {
-                    error_log($mysql_error($db));
+                    error_log(mysql_error($db));
                     echo "There was an error setting the flag.";
                 }
                 $flags = $newflags;
@@ -1289,12 +1289,12 @@ else if (isset($_REQUEST['users'])) {
         if ($rebuild || $sortTitle == "")
             $setVars[] = "sort_title = '"
                          . mysql_real_escape_string(strtoupper(
-                             getSortingTitle($title))) . "'";
+                             getSortingTitle($title)), $db) . "'";
 
         if ($rebuild || $sortAuthor == "")
             $setVars[] = "sort_author = '"
                          . mysql_real_escape_string(strtoupper(
-                             getSortingPersonalNameList($author))) . "'";
+                             getSortingPersonalNameList($author)), $db) . "'";
 
         $setVars = implode(",", $setVars);
         $result = mysql_query(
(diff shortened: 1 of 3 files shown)


www/adminops
```diff
@@ -1289,12 +1289,12 @@ else if (isset($_REQUEST['users'])) {
         if ($rebuild || $sortTitle == "")
             $setVars[] = "sort_title = '"
                          . mysql_real_escape_string(strtoupper(
-                             getSortingTitle($title))) . "'";
+                             getSortingTitle($title)), $db) . "'";
```
dfabulich, line 1292: I tested this feature locally by going to http://localhost:8080/adminops and clicking "Rebuild ALL author/title sort keys in GAMES table" http://localhost:8080/adminops?fixsortkeys&rebuild

In main, it didn't work, in the branch, it did. (But I have no idea what it does, so… 🤷‍♂️)

www/adminops
```diff
@@ -1849,7 +1849,7 @@ else if (isset($_REQUEST['users'])) {
                filterID, filterName, ckBoxName, showName, endDate,
                filterType, explanation
              from filters
-             where filterid = '$qFilterID'");
+             where filterid = '$qFilterID'", $db);
```
dfabulich, line 1852: I was unable to test this feature locally. It looks like it ought to work if you go to http://localhost:8080/adminops and click on "View/edit game filter list" http://localhost:8080/adminops?filters but if you go there, it says, "Table 'ifdb.filters' doesn't exist" 

And, indeed, that's not a table, and never has been since our first data drop from MJR in January, so I think this is just dead code.

salty-horse: Should I delete everything related to that query in a separate commit?

www/util.php
```diff
@@ -2915,7 +2915,7 @@ function close_user_acct($db, $uid, $stat, &$progress)
 
     // unlock tables, if we haven't already
     if ($tableLocks)
-        mysql_query("unlock tables");
+        mysql_query("unlock tables", $db);
```
dfabulich, line 2918: This line is supposed to be unreachable; it would only be reached if there were errors after locking that prevented unlocking.

www/adminops
dfabulich: Eh, it's certainly not the only dead code in the code base. If you went on a dead-code hunt as a separate PR, that would be welcome

www/util.php
dfabulich: I found a way to test it by adding a tag as test@ifdb.org then deliberately adding a SQL syntax bug on line 2799, and trying to delete test@ifdb.org. In main, the deletion fails and the result page is blank. In your branch, it works (displaying an error message) as intended.

dfabulich:
Thanks!

PR #163, opened 2021-05-28
Silence static analyzer errors by adding missing includes, moving functions around

salty-horse:
1. Some files were missing explicit includes (These are already included by including other files).
2. Some helper functions and defines were in the wrong scope.
3. `base_convert` expects a string, so added an explicit conversion from int.

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -7,7 +7,7 @@ include_once "login-persist.php";
 $userid = checkPersistentLogin();
 
 include_once "util.php";
-include "pagetpl.php";
+include_once "pagetpl.php";
 include_once "useractivation.php";
 
 include_once "dbconnect.php";
--- a/www/adminops
+++ b/www/adminops
@@ -1601,18 +1601,6 @@ else if (isset($_REQUEST['users'])) {
     echo "<p>Rows successfully updated: $okCnt<br>";
 
 } else if (isset($_REQUEST['cleanpix'])) {
-
-    function adjustImageName($name)
-    {
-        if (isLocalDev()
-            && strpos($name, ":") !== false)
-        {
-            $name = explode(":", $name);
-            $name[0] = "0";
-            $name = implode(":", $name);
-        }
-        return $name;
-    }
     echo "<h1>Scanning for unreferenced images</h1>"
         . "<form name=\"fixpix\" method=post action=\"adminops\">"
         . "<input type=hidden name=cleanpix value=1>";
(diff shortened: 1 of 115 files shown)


dfabulich (changes requested):
I get a 500 error loading all pages when I use this PR. The Docker log says:

```PHP Fatal error:  Cannot redeclare basePageHeader() (previously declared in /var/www/html/pagetpl.php:6) in /var/www/html/pagetpl.php on line 6```

dfabulich:
(Sorry for taking so long to review this!)

salty-horse:
Sorry, I'll check it out.

salty-horse:
The issue was that the two additional includes I added in `util.php` were included by other files using `include` instead of `include_once`.
I fixed those but I'm worried that I may have missed another double-include situation.

I wonder if I should change more includes to be on the same side. There are now 32 remaining uses of `include` instead of `include_once`, that's used to include files with utility functions, but some of those have code outside of functions (like `www/imageuploadhandler.php`, and may have a good reason to be included twice...?

dfabulich:
After some Googling on this? I think all of our includes should actually be using `require_once`. Maybe as a separate PR?

salty-horse:
Do you mean using the hard-error `require_once` instead of `include` and `include_once`?

dfabulich:
Yeah. Right now, we're using a lot of `include`, and all of those should really be `include_once`, _and_ they should really be `require`, and therefore they should be `require_once`.

In practice, I don't think the include vs require point matters much (basically none of these pages will work at all if their inclusions are missing) but if we're gonna update them in bulk, seems like that's what we should update them to.

salty-horse:
I can do that. What does this mean to this PR?

dfabulich:
I think I misunderstood the conversation we had about this PR. If you think it’s ready to merge, I’ll review it again and merge it.

salty-horse:
I rebased my changes onto `main`. I didn't try re-running the server, so if you can perform any basic tests that would be cool.

dfabulich:
As I read this, the only localized spots this could have broken would be in searches and in `generateTUID`.

I tested searches (`isNameSuffix`, `unNegate`) by searching for `author:Xalavier Nelson Jr.` and again with `-screw author:Xalavier Nelson Jr.` and it worked.

I tested `generateTUID` by testing creating a game.

Clicking around, I found no other issues.

PR #164, opened 2021-05-28
Fix SQL injection in editlist

salty-horse:
As mentioned on Discourse, user input wasn't properly escaped in some situations.
I think it's pretty harmless. The most is can do is cause a checker query to run a bit slower and return a "yes" instead of a "no".

### DIFF
--- a/www/editlist
+++ b/www/editlist
@@ -253,7 +253,8 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST'
                 $deferred = true;
             } else {
                 // we have a resolution - set the resolved TUID in the item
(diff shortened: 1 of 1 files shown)


dfabulich:
I'm merging this because it looks safe and it's been a month already, but I can't figure out for the life of me how to do a regression test of this code. I tried creating a list, adding a few entries, removing entries from the list, and deleting the list, but nothing I did triggered the code you're modifying.

I hypothesize that it's dead code and would be better off deleted, but I'm not confident enough to prove that right now. Specifically, I think `editlist` at one point had some half-hearted game search functionality, which has since been replaced by the game search popup.

PR #169, opened 2021-08-06
Make "Development System" field on game details page clickable

NilsFagerburg:
The same way the author field works.


dfabulich:
Sorry for taking so long to look at this! I'm merging it as is because it mostly works, but while testing it, I did turn up some issues:

https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/317

https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/316

PR #171, opened 2021-10-09
Load recommendations on the home page asynchronously

dfabulich:
The recommendation engine can take seconds to run. It starts by querying all of the games that the current user has rated, and then finds all of the users that have rated those games, then finds all of their ratings. It stores the results in a temporary table to find users who like the same games that you like ("like-minded users"). It then attempts to find games that you haven't rated that are highly rated by the group, and if that fails, it attempts to find games that any like-minded user likes.

That's… a lot.

In this implementation, the "IFDB Recommends" section now loads asynchronously if you're logged in and we're computing recommendations, replacing the "IFDB Recommends" section with "Loading..." and using JavaScript to make a request for recommendations in the background. When the recommendations are computed, we inject them directly into the page, and cache them in a server-side session cache, so refreshing the home page in the future will show recommendations directly in the HTML source.

### DIFF
--- a/www/async-recommendations
+++ b/www/async-recommendations
@@ -1,0 +1,19 @@
+<?php
+include_once "session-start.php";
+include_once "dbconnect.php";
+include_once "util.php";
+include "pagetpl.php";
+include_once "login-persist.php";
+include "newitems.php";
+include_once "commentutil.php";
+
+$db = dbConnect();
+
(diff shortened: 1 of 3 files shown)


curiousdannii:
Such a small commit! An advantage of server-side rendering I guess.

PR #180, opened 2021-11-17
Move images to img subdirectory

Iniquit:
This relatively unglamorous PR declutters IFDB's `www` directory by kicking ~120 image files over into a new `/img` subdirectory. Most of the remaining ones don't _appear_ to be used anywhere in the source and might be candidates for deletion.

I reviewed most of these relocations manually after the initial find-and-replace but this could still use more pairs of eyes on it to make sure I didn't break any of the links.


Iniquit:
https://github.com/iftechfoundation/ifdb/blob/92237311fac6883234badf21527406b42116d253/www/home#L81
I actually caught one omission already-- the "Show me how!" button in the "New to IF?" section on the home page.  `img src="dlhelpbtn.gif"`  should be `img src="img/dlhelpbtn.gif"`.

dfabulich:
This caused https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/326

PR #191, opened 2021-12-06
Re-add Home link for custom themes who hide the header

dfabulich:
The Home link is still hidden at mobile sizes, which I *think* is what we want, but I'm not sure…

Deployed to https://dev.ifdb.org/ 

@Iniquit


lkcampbell (commented):
Approved.

PR #199, opened 2021-12-09
Make latest simplest viewgame version be the rel=canonical

dfabulich:
This will send a hint to search engines, suggesting that they prevent history variants from showing up in search results.

https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -644,6 +644,9 @@ if ($hasart) {
     $extraHead .= get_root_url() . "viewgame?coverart&amp;id=$id\"/>";
 }
 
+$extraHead .= "<link rel=\"canonical\" href=\"";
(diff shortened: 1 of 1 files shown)


lkcampbell (approved):
The changes look fine based on the article you linked.

How are we testing if it works? Is there even a way to do that?

dfabulich:
I can use Google Search Console to make sure Google can see/read the hint. Google isn't required to honor the hint in any case.

dfabulich:
(This came up in the context of an author who wanted to change their author name, but search engines were indexing the historical v=1 variation. I'll request that Google specifically recrawl that page when this deploys.)

PR #202, opened 2021-12-16
Fix nav header on pages that include login-persist.php before util.php

dfabulich:
The home page, the search page, and others all have an incorrect nav bar after logging in. (The bar says "Log In" when it should say "Profile, Settings, My Activity, Inbox, Logout.") I think this is a regression from my nav header changes PR #193.

PHP does something weird when you introduce circular dependencies; if A includes B and B includes A, it runs B right up until the point where B includes A, then runs A to completion, and then runs B to completion. That's no good when the dependency is something essential for A to run correctly.

So in this PR we're making util.php depend on session-start.php and ensuring that session-start.php depends on nothing, which guarantees that this will be safe.

### DIFF
--- a/www/session-start.php
+++ b/www/session-start.php
@@ -1,6 +1,19 @@
 <?php
 
-include_once "util.php";
+define("PRODUCTION_SERVER_NAME", "ifdb.org");
+define("STAGING_SERVER_NAME", "dev.ifdb.org");
+
+function isProduction() {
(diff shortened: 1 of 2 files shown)


dfabulich:
This looks pretty bad, so I'm just merging and deploying it now.

lkcampbell:
I saw this problem  when the change first went in but I just did a hard reload and it went away.

PR #205, opened 2021-12-21
Remove Edit Privileges checkboxes from adminops user editor

dfabulich:
If we want to grant individuals access to edit formats and operating systems, we can just grant them access to all formats or all OSes. (And, most likely, only admins will ever do it.)

screenshot

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -934,26 +934,6 @@ else if (isset($_REQUEST['users'])) {
         $priv[] = $privMap[$privcode{$c}];
     $privNames = implode(", ", $priv);
 
-    $result = mysql_query(
-        "select id, fmtname, userid
-         from filetypes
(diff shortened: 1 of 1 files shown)


lkcampbell:
@dfabulich I see you removed the queries and the mass of checkboxes.

I also see you changed the Login history layout. Was this intentional? I only ask because you didn't mention it in your description.

dfabulich:
Ah, yeah, good catch. When I first just deleted the "Format Editing Privileges" and "OS Privileges" section, the page looked like this, with the "Login History" section pushing down the "impersonate" link and the Apply button, so I moved it out to the side bar (and made it wrap inline on mobile).

broken-screenshot

lkcampbell (approved):

PR #209, opened 2021-12-21
Exclude editorial reviews from member reviews

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/261

I rebased this into two commits, one of which reformats the queries, adding whitespace so you can actually read them, and doing the actual work in a second commit with a readable diff.


dfabulich:
This newly rebased version of the PR should be much easier to review.

lkcampbell (approved):

PR #234, opened 2023-01-22
Anonymous polls and Locked polls

dfabulich:
In an effort to support the upcoming IFDB Awards, we're introducing a new feature to make polls anonymous, and to allow users to lock polls for voting (so an official winner can be declared).

In this PR, we create a new `flags` column on the polls table, and define the `1` flag to represent the first flag, the `POLL_FLAG_ANONYMOUS`, and the `2` flag to be `POLL_FLAG_LOCKED`. When creating a poll, you can choose whether to make the poll "Public" (which is the default) or "Anonymous."

image

# Here's an example of what an anonymous poll looks like when there's been a vote:

image

Note that users may still type a "quick quote" and notes, but they won't be visible.

# IFDB Administrators will see an "Admin: View all votes" link.

Screenshot 2023-01-22 at 8 11 14 PM

When you click it, it will take you to the poll page with "&unlock=1" in the URL, and then all votes will be shown to the administrator.

# Set a poll to be locked

image

# Viewing a locked poll

image

#Trying to vote in a locked poll in a separate window

In this example, I opened the poll in two tabs. In the first tab, I viewed the poll while it was unlocked, then, in the second tab, I locked the poll. Back in the first tab, I attempted to submit the vote, and got an appropriate error message.

image

Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/373

This code is now live on dev.ifdb.org

### DIFF
--- a/sql/patch-schema.sql
+++ b/sql/patch-schema.sql
@@ -1,5 +1,10 @@
 USE ifdb;
 
+-- Jan 2022
+ALTER TABLE `polls`
+  ADD COLUMN `flags` int(11) NOT NULL DEFAULT 0
+;
+
 DROP TABLE IF EXISTS `audit`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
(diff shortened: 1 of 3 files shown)


brirush84:
Thanks!

I expect that the code is likely bad; I assumed that you would be able to make something much better, but I was just trying to prove a point (a few people were suggesting that author's choice might not be able to run this year because making polls anonymous would be too hard). So if it doesn't work, feel free to trash it. If it does work but only needs some slight tweaks, I could work on them.

PR #260, opened 2023-10-19
PHP 8 fixes

salty-horse:
Initial work on iftechfoundation/ifdb#468.

The site seems to work - I didn't test all features, but browsing around I didn't encounter any problems.

Based on the SQL error, I assume MySQL was also upgraded, so there might be errors in the some of the commands I didn't run.

I ran `php -l` on all the files and fixed the warnings. Some of these commits can be applied to the main (php7) branch as well, if you wish.

Since there's no test suite, is it possible to deploy this on to a test site with some coverage analysis? Or is that too resource-consuming on a live site?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7-apache
+FROM php:8-apache
 
 RUN docker-php-ext-install mysqli
 RUN apt-get update -y && apt-get install -y zlib1g-dev libpng-dev libjpeg-dev
--- a/www/util.php
+++ b/www/util.php
@@ -52,54 +52,6 @@ function mysql_result($result, $row, $field = 0) {
 }
 function mysql_insert_id($linkid) { return mysqli_insert_id($linkid); }
 
-/**  
- * Polyfill for mysqli_execute_query, available in PHP 8.2
- * Copied from https://php.watch/versions/8.2/mysqli_execute_query
- *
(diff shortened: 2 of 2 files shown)


dfabulich:
I'll figure out how to deploy it!

dfabulich:
I haven't forgotten this! I think the thing I have to do is to Dockerize IFDB in production, which we currently don't do. (We only Dockerize in the dev environment.) This will allow us to control our PHP version, independent from whatever Debian tries to foist on is.

I just need to get my butt in chair and figure out how best to do that. In this case, that's going to mean porting the DB from a non-Dockerized DB into a Dockerized DB, which I personally have never done before, so this will require some research.

salty-horse:
Most of these are applicable to the current setup, so I can commit them.

salty-horse:
@dfabulich Can we test this? I can create a separate PR with the stuff that's compatible with both 7 and 8. (One fix was already applied in the mentioned PR.)

dfabulich:
> I can create a separate PR with the stuff that's compatible with both 7 and 8.

That sounds like a great idea.

salty-horse:
I rebased to `main`, since the other commits are in, and changed this PR to merge to `main` instead of `php-8`.

salty-horse:
This is no longer relevant (and I accidentally deleted `MAX_PASSWORD_LENGTH`)

PR #267, opened 2023-11-24
Replace "what is a wiki" links

salty-horse:
The main wiki at http://wiki.org/ isn't online, and I found no news about it, so for now the links can point to Wikipedia's explanation of the term "Wiki".


dfabulich:
Thanks! This is now deployed.

PR #269, opened 2023-12-27
Use gametags table when searching for tags

salty-horse:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/402.

This change takes all the `tag:FOO` parts of a search query and turns them into a subselect from the gametags table, which has an index.

Notes:
* The existing syntax didn't allow for the `+` and `-` prefixes, and assumed all tags are AND-ed. I preserved it the only way I know how: with a join. If there's a nicer way, please share. (I found [this trick based on the sum of tag IDs](https://dba.stackexchange.com/questions/226670/select-items-that-has-one-or-more-specific-tags) but it only works if all the tags are in a separate table and have a unique numerical ID - that's not the case).
* Because I might be building a big subselect string, I limited the number of tags you can search by. I think it's sensible, and probably no one will reach that limit or notice their extra tag requirements are ignored.
* Is it better to make this a join instead of a subselect?
* The subselect is `games.id in (select t0.gameid from gametags ..)`. Should it use `distinct t0.gameid` to remove the duplication of records that can happen if many users reported a tag? Is there any benefit to it, and is it worth the cost?
* I noticed that sandboxed users' tags aren't ignored. Bug? Abusers can spam the tags for games they don't like.
* Querying with special keywords like "tags" in combination with regular terms works unexpectedly. "tag:foo bar" looks for a tag called "foo bar" instead of games named "bar" with tag "foo". Annoying, but maybe someone depends on it already.

Some queries I tested this with:
* `tag:fantasy` - single tag.
* `tag:parser tag:humor tag:fantasy` - multiple tags.
* `tag:adjustable difficulty tag:achievements` - multiple tags with spaces in name.
* `dragon tag:IFComp 2020` - tag and keyword.

I didn't perform any measurements. I'm assuming this works faster than the current `RLIKE` query based on the results. I did paste the query into a mysql shell with `EXPLAIN` but I'm not an expert at reading it.

### DIFF
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -989,7 +989,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
         $having = "having $having";
 
     // Build tags subselect
-    $tagsSubselect = "";
+    $tagsTable = "";
     if ($tagsToMatch) {
         // Limit number of tags to avoid abuse
         $tagsToMatch = array_slice($tagsToMatch, 0, 20);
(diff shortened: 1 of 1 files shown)


salty-horse:
I changed it to use a join, in a separate commit. You can decide which is better.

Based on tests, I think I'm correct in my assumption that in the current behavior, searching for multiple tags `AND`s them. If that's the case, what's the point of each result having a label of "Tags matched: A B C"? Why not have those links at the top of the list?
tags_matched

salty-horse:
@dfabulich What should be done about the "Tags Matched" label, and the issue I raised about sandboxed users having visible tags?

dfabulich:
I've filed https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/422 on "Tags Matched" and https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/423 on sandboxed tags.

PR #270, opened 2024-01-02
Generate a materialized view for `gameRatingsSandbox0`

dfabulich:
Now we'll use `gameRatingsSandbox0_mv` for pretty much all purposes.

Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/401 and fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/407

### DIFF
--- a/sql/ifdb-to-ifarchive-scrubber.sql
+++ b/sql/ifdb-to-ifarchive-scrubber.sql
@@ -45,6 +45,12 @@ update users
 drop view gameRatingsSandbox0;
 drop view gameRatingsSandbox01;
 
+drop table gameRatingsSandbox0_mv;
+drop procedure refresh_gameRatingsSandbox0_mv;
+drop trigger reviews_insert;
+drop trigger reviews_update;
+drop trigger reviews_delete;
+
 /* delete troll users and their reviews and comments */
 delete from ucomments
   where userid in (select id from users where sandbox = 1);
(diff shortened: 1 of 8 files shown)


salty-horse:
Is `embargopastdate` supposed to be used anywhere? I assume it's in the definition of the view that's not part of the repository.

Perhaps both the view and the cron jobs should be documented somewhere.

dfabulich:
I added more comments about `embargopastdate` to `refresh-embargoed-reviews`.

```
// This URL will be pinged on a regular schedule via a cron job.

// This cron job is really strange.

// The purpose of the cron job is to keep the `gameRatingsSandbox0_mv`
// materialized view up to date as time passes.

// The `gameRatingsSandbox0` view automatically excludes reviews with an
// embargo date in the future; thus, the view is like a `join` of the
// `reviews` table and the current datetime.

// But the trigger that keeps `gameRatingsSandbox0_mv` updated only runs
// when we `update` the `reviews` table. The trigger doesn't naturally
// update the MV as time passes.

// This cron job incorporates the current datetime into the materialized
// view by translating the passage of time into `update` statements on the
// `reviews` table.

// The `reviews` table has an `embargopastdate` column, which serves no
// real purpose other than to be a column that we can update in order to
// activate the MV update trigger. We only read `embargopastdate` here,
// in this query, just in order to find recently unembargoed reviews that
// we haven't incorporated into the MV.

// It's kind of like running `touch` on a file, updating its lastmodified
// date, in order to convince `make` to do something with it. We're
// "touching" the `embargopastdate` in order to activate the post-update
// trigger, refreshing the MV.
```

PR #271, opened 2024-01-02
Use fulltext index when searching by author

dfabulich:


dfabulich:
@salty-horse  This doesn't merge cleanly with #269. Do you have time to rebase it onto that?

salty-horse:
Let's do these one by one and only rebase on final work.

I suggested this as kind of a hack to use the existing fulltext index. Is this preferable to creating a new index just for the author name? What happens when there are two separate fulltext indexes on different groups of columns? Do they share some of the data, making it a cheap choice, or is it separate?

My other worry is some comments suggesting that the query plan will sometimes opt to not use the fulltext index when there's a mix of other conditions, but I don't know if it's relevant.

dfabulich:
ah, yes, I understand it now. As far as I can tell, indexes are cheap for us, and CPU is expensive, so let's just create more indexes.

PR #275, opened 2024-01-03
Rollback implicit TUID search

dfabulich:
Rolling back implicit TUID search is necessary to fix a performance issue https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/404

We're also updating the documentation to make it clear how to search by IFID or TUID in the search box, addressing https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/300.

### DIFF
--- a/www/gamesearchpopup.php
+++ b/www/gamesearchpopup.php
@@ -197,9 +197,9 @@ function gameSearchPopupDiv()
           </div>
        </div>
        <div class="edit-popup-win">
-          <p><b>Step 1:</b> Search for a game by title, <?php
-              echo helpWinLink("help-ifid", "IFID") ?>, or <?php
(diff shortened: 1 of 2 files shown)


salty-horse:
I don't like those search instructions - they assume knowledge of the search syntax. There should be a clearer example on how to search by TUID/IFID beyond just mentioning the prefix. Maybe it's enough to change `tuid:` to `tuid: 4gxk83ja4twckm6j `, and `ifid:` to `ifid:ZCODE-12-990623-CBCB`. (Using Zork as an example.)

I still prefer a "do what I mean" approach, like I mentioned in the linked comment. Improved idea:
If the search query includes a single term without a prefix, the WHERE can be `id = X OR <ifid search> OR <regular word search>`.

But this can be a task for later.

dfabulich:
Rewriting this as #278

PR #277, opened 2024-01-03
Add fulltext index for author search

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/406

### DIFF
--- a/sql/patch-schema-2.sql
+++ b/sql/patch-schema-2.sql
@@ -2,6 +2,8 @@ USE ifdb;
 
 -- use this script for pending changes to the production DB schema
 
+alter table games add fulltext key `author` (`author`);
+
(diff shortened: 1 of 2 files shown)


salty-horse:
In the other commits, I enabled the use of bind parameters instead of calling an escape function. It will be cumbersome to switch to  it without committing the other stuff, so can be done later.

In the closed https://github.com/iftechfoundation/ifdb/pull/271/ I also wrote a better (I think) sanitation function for match expressions: `removeSqlMatchChars`. I don't know if you want to use it. Without it, users can make more flexible matches with special characters, but it might have surprising behaviors since it's not documented. (It's not great, since it strips things like `-`.)

PR #276, opened 2024-01-03
Remove LIKE clauses from fulltext search, to reactivate fulltext index

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/404

My naive implementation of implicit TUID search caused a performance issue, and it was never working with implicit IFID searches anyway, https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/300

Now we're using a fast mechanism for implicitly searching by TUID and IFID for one-word searches, and letting the `match` query match exactly.

Intuitively, it really seems like this ought to have broken something… but, as far as I can tell, all the searches seem to behave exactly the same as before. I'm not sure I understand why yet.

I explained my approach here: https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/404#issuecomment-1874844579

> Starting with the `game-ratings-materialized-view` branch, I tried a simple search for Zork `http://localhost:8080/search?searchbar=Zork`.
> 
> It ran this query:
> 
> ```sql
> select
>     games.id as id,
>     games.title as title,
>     games.author as author,
>     games.tags as tags,
>     games.moddate as moddate,
>     games.system as devsys,
>     if (time(games.published) = '00:00:00',
>         date_format(games.published, '%Y'),
>         date_format(games.published, '%M %e, %Y'))
>         as pubfmt,
>     if (time(games.published) = '00:00:00',
>         date_format(games.published, '%Y'),
>         date_format(games.published, '%Y-%m-%d'))
>         as published,
>     date_format(games.published, '%Y') as pubyear,
>     (games.coverart is not null) as hasart,
>     avgRating as avgrating,
>     numRatingsInAvg as ratingcnt,
>     stdDevRating as ratingdev,
>     numRatingsTotal,
>     numMemberReviews,
>     games.sort_title as sort_title,
>     games.sort_author as sort_author,
>     ifnull(games.published, '9999-12-31') as sort_pub,
>     games.flags
>     , match (title, author, `desc`, tags) against ('Zork' in boolean mode) as relevance
> from
>     games
>     left join gameRatingsSandbox0_mv on games.id = gameid
> where
>     ( (match (title, author, `desc`, tags) against ('Zork' in boolean mode) or (title like 'Zork'  or ( title like '%Zork%'))))
> 
> order by
>     if(title = 'Zork',0,1),if(title like 'Zork%',0,1),if((title like 'Zork'  or ( title like '%Zork%')),0,1),relevance desc
>     
> limit 0, 20;
> ```
[...]

PR #278, opened 2024-01-03
Implicit TUID IFID search

dfabulich:
My naive implementation of implicit TUID search caused a performance issue https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/404

… and it was never working with implicit IFID searches anyway, https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/300

Now we're using a fast mechanism for implicitly searching by TUID and IFID for one-word searches.


dfabulich:
Actually, I'm just going to bake this into #276

PR #283, opened 2024-01-03
Don't update `reviews.moddate` when updating `embargopastdate`

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/428

### DIFF
--- a/sql/patch-schema-2.sql
+++ b/sql/patch-schema-2.sql
@@ -8,13 +8,15 @@ alter table reviews
     add key `embargopastdate` (`embargopastdate`)
 ;
 
-update reviews
-set embargopastdate = now()
+update reviews r1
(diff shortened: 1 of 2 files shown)


dfabulich:
This bug is messing up `moddate` for a bunch of reviews, so I'm just going to merge it now.

salty-horse:
I don't understand what setting `moddate` to its own value is doing.

dfabulich:
`moddate` automatically updates itself to `now()`, unless you manually set it to something. Setting it to its own old value prevents it from changing, which is what we need in this case.

salty-horse:
Maybe a comment explaining it would help...? Unless it's obvious and I'm silly.

PR #285, opened 2024-01-04
Fetch negated tags from gametags

salty-horse:
Proper fix for https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/426

### DIFF
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -400,6 +400,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
     $having = "";
     $relevance = "";
     $tagsToMatch = [];
+    $tagsToNegate = [];
 
     // add in the full-text part, if applicable
     if (count($words))
@@ -552,9 +553,8 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
         case 5:
             // tags
             if ($negate) {
-                $expr = "tags rlike '[[:<:]]"
-                    . mysql_real_escape_string(quoteSqlRLike($txt), $db)
-                    . "[[:>:]]'";
+                $tagsToNegate[] = $txt;
+                $negate = false; // Turn off negate flag as tags are handled separately
             } else {
                 $tagsToMatch[] = $txt;
             }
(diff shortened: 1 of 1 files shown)


dfabulich:
I just noticed https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/432 which this doesn't fix, but it seems like maybe it ought to.

salty-horse:
It's impossible to use a join for the negative tags, so I'm fetching them in a subquery.

My initial inclination was to write it like this:
> and gameid NOT IN (select distinct gameid from gametags where tag in (list of tags to negate)

However, I read recommendations to do this:
> and NOT EXISTS (select 1 from gametags where games.id = gameid and tag in (list of tags to negate))

It's a bit counter-intuitive to me as I'm not sure how the `games.id` binds to the outer one, which isn't a single value, and what kind of operation it translates to internally, but this is in [official MySQL examples](https://dev.mysql.com/doc/refman/8.0/en/exists-and-not-exists-subqueries.html). (MariaDB doesn't have a similar document - I assume it's the same.)

I'm also not certain about the performance of the first option vs the second.

Analyzing this tells me it uses the gametags index in all cases:
```
select games.id as id, games.title 
from games 
join (select distinct t0.gameid 
      from gametags as t0 join gametags as t1 
      where t0.tag = "fantasy" and t1.tag = "twine") as gt on games.id = gt.gameid 
where 
  not exists ( select 1 from gametags where games.id = gameid and tag in ("horror"));

+------+--------------+------------+------+---------------+------+---------+---------------+-------+----------+----------+------------+----------------------------------------+
| id   | select_type  | table      | type | possible_keys | key  | key_len | ref           | rows  | r_rows   | filtered | r_filtered | Extra                                  |
+------+--------------+------------+------+---------------+------+---------+---------------+-------+----------+----------+------------+----------------------------------------+
|    1 | PRIMARY      | games      | ALL  | PRIMARY       | NULL | NULL    | NULL          | 11017 | 11017.00 |   100.00 |      96.08 | Using where                            |
|    1 | PRIMARY      | <derived2> | ref  | key0          | key0 | 34      | ifdb.games.id | 1     | 0.06     |   100.00 |     100.00 |                                        |
|    3 | MATERIALIZED | gametags   | ref  | gameid,tag    | tag  | 257     | const         | 336   | 460.00   |   100.00 |     100.00 | Using index condition                  |
|    2 | DERIVED      | t0         | ref  | tag           | tag  | 257     | const         | 425   | 654.00   |   100.00 |     100.00 | Using index condition; Using temporary |
|    2 | DERIVED      | t1         | ref  | tag           | tag  | 257     | const         | 1940  | 1.00     |   100.00 |     100.00 | Using where; Using index; Distinct     |
+------+--------------+------------+------+---------------+------+---------+---------------+-------+----------+----------+------------+----------------------------------------+
```

I tested this with:
* `tag:vampire tag:fantasy tag:twine -tag:horror`
* `dragon tag:fantasy  -tag:horror tag:twine -tag:vampire -tag:dragon` (finding games mentioning "dragon", but not tagged with "dragon")

BTW in my tests I tripped several times by placing a regular keyword after a special-prefixed one, and it getting gobbled up into the special prefix. Still finding it unexpected even after I learned about it.

PR #291, opened 2024-01-10
Remove implicit search by IFID

dfabulich:
We added implicit search by IFID in https://github.com/iftechfoundation/ifdb/pull/276 to fix https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/300 ("Can't select game by TUID or IFID when editing Cross-References ")

But this caused https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/439 and https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/433

IMO, implicit search by IFID just isn't worth it, especially when implicit search by TUID works great.

There are a few places that do support implicit search by TUID and IFID still, including polls and "Suggest Similar Games" at  `/crossrec`. They still work fine, and I'm not going to bother to rip anything out there. 

(BTW, note the difference between "cross-recommendations", which is for "similar games", and "cross-references", which is for games that are formally related to each other in an enumerated way: adaptations, ports, sequels, translations, etc.)

### DIFF
--- a/www/gamesearchpopup.php
+++ b/www/gamesearchpopup.php
@@ -197,8 +197,7 @@ function gameSearchPopupDiv()
           </div>
        </div>
        <div class="edit-popup-win">
-          <p><b>Step 1:</b> Search for a game by title, <?php
-              echo helpWinLink("help-ifid", "IFID") ?>, or <?php
+          <p><b>Step 1:</b> Search for a game by title or <?php
               echo helpWinLink("help-tuid", "TUID") ?>:
           <br>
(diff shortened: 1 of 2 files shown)


salty-horse:
Would it useful to retain the automatic IFID search if it's in a known format, such as UUID, `(ZCODE|GLULX)-\d+-\d+-[0-9a-z]+`, etc.?
Maybe too much work for a convenience feature, as each system seems to generate its own IFID format.

salty-horse (approved):

PR #293, opened 2024-01-11
Revert "Add fulltext index for author search"

dfabulich:
This reverts https://github.com/iftechfoundation/ifdb/pull/277, which caused https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/436

With the benefit of hindsight, nobody should have trusted that #277 did what it was supposed to do, because it doesn't include the results of before/after `analyze`. It turns out that it didn't fix https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/406 at all; instead, it just caused https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/436, for no performance benefit.

This fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/436 but does not address https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/168 which is also about author search relevance.

Before reverting:
```sql
select sql_calc_found_rows
    games.id as id,
    games.title as title,
    games.author as author,
    games.tags as tags,
    games.moddate as moddate,
    games.system as devsys,
    if (time(games.published) = '00:00:00',
        date_format(games.published, '%Y'),
        date_format(games.published, '%M %e, %Y'))
        as pubfmt,
    if (time(games.published) = '00:00:00',
        date_format(games.published, '%Y'),
        date_format(games.published, '%Y-%m-%d'))
        as published,
    date_format(games.published, '%Y') as pubyear,
    (games.coverart is not null) as hasart,
    avgRating as avgrating,
    numRatingsInAvg as ratingcnt,
    stdDevRating as ratingdev,
    numRatingsTotal,
    numMemberReviews,
    games.sort_title as sort_title,
    games.sort_author as sort_author,
    ifnull(games.published, '9999-12-31') as sort_pub,
    games.flags
    
from
    games
                left join gameRatingsSandbox0_mv on games.id = gameid
    
where
    ( match (author) against ('Emily Ryan')  or author rlike '[[:<:]]E.*[[:space:]]+Ryan[[:>:]]')
    


order by
    starsort desc
    
limit 0, 20;
```

```
+------+-------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
| id   | select_type | table                  | type   | possible_keys | key     | key_len | ref           | rows  | r_rows   | filtered | r_filtered | Extra                                        |
+------+-------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
[...]

PR #294, opened 2024-01-13
Bring back exact substring author search

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/436

`match against` is not strict, and author search never sorts by relevance, so we need to bring back the like clause for exactness.

With the benefit of hindsight, nobody should have trusted that #277 did what it was supposed to do, because it doesn't include the results of before/after `analyze`. It turns out that it didn't fix https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/406 at all; instead, it just caused https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/436, for no performance benefit.

This fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/406 and https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/436 but does not address https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/168 which is also about author search relevance.

Putting the `like` and `rlike` clauses in an `AND` subclause allows us to actually use the author index.

Before reverting:
```sql
select sql_calc_found_rows
    games.id as id,
    games.title as title,
    games.author as author,
    games.tags as tags,
    games.moddate as moddate,
    games.system as devsys,
    if (time(games.published) = '00:00:00',
        date_format(games.published, '%Y'),
        date_format(games.published, '%M %e, %Y'))
        as pubfmt,
    if (time(games.published) = '00:00:00',
        date_format(games.published, '%Y'),
        date_format(games.published, '%Y-%m-%d'))
        as published,
    date_format(games.published, '%Y') as pubyear,
    (games.coverart is not null) as hasart,
    avgRating as avgrating,
    numRatingsInAvg as ratingcnt,
    stdDevRating as ratingdev,
    numRatingsTotal,
    numMemberReviews,
    games.sort_title as sort_title,
    games.sort_author as sort_author,
    ifnull(games.published, '9999-12-31') as sort_pub,
    games.flags
    
from
    games
                left join gameRatingsSandbox0_mv on games.id = gameid
    
where
    ( match (author) against ('Emily Ryan')  or author rlike '[[:<:]]E.*[[:space:]]+Ryan[[:>:]]')
    


order by
    starsort desc
    
limit 0, 20;
```

```
+------+-------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
| id   | select_type | table                  | type   | possible_keys | key     | key_len | ref           | rows  | r_rows   | filtered | r_filtered | Extra                                        |
+------+-------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
|    1 | SIMPLE      | games                  | ALL    | NULL          | NULL    | NULL    | NULL          | 12987 | 12987.00 |   100.00 |       1.00 | Using where; Using temporary; Using filesort |
|    1 | SIMPLE      | gameRatingsSandbox0_mv | eq_ref | PRIMARY       | PRIMARY | 34      | ifdb.games.id | 1     | 0.91     |   100.00 |     100.00 |                                              |
+------+-------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
```

After:

```sql
select sql_calc_found_rows
    games.id as id,
    games.title as title,
    games.author as author,
    games.tags as tags,
    games.moddate as moddate,
    games.system as devsys,
    if (time(games.published) = '00:00:00',
        date_format(games.published, '%Y'),
        date_format(games.published, '%M %e, %Y'))
        as pubfmt,
    if (time(games.published) = '00:00:00',
        date_format(games.published, '%Y'),
        date_format(games.published, '%Y-%m-%d'))
        as published,
    date_format(games.published, '%Y') as pubyear,
    (games.coverart is not null) as hasart,
    avgRating as avgrating,
    numRatingsInAvg as ratingcnt,
    stdDevRating as ratingdev,
    numRatingsTotal,
    numMemberReviews,
    games.sort_title as sort_title,
    games.sort_author as sort_author,
    ifnull(games.published, '9999-12-31') as sort_pub,
    games.flags
    
from
    games
                left join gameRatingsSandbox0_mv on games.id = gameid
    
where
    ( match (author) against ('Emily Ryan') and (author like '%Emily Ryan%'  or author rlike '[[:<:]]E.*[[:space:]]+Ryan[[:>:]]'))
    


order by
    starsort desc
    
limit 0, 20;
```

```
+------+-------------+------------------------+----------+---------------+---------+---------+---------------+------+--------+----------+------------+----------------------------------------------+
| id   | select_type | table                  | type     | possible_keys | key     | key_len | ref           | rows | r_rows | filtered | r_filtered | Extra                                        |
+------+-------------+------------------------+----------+---------------+---------+---------+---------------+------+--------+----------+------------+----------------------------------------------+
|    1 | SIMPLE      | games                  | fulltext | author        | author  | 0       |               | 1    | 130.00 |   100.00 |       0.77 | Using where; Using temporary; Using filesort |
|    1 | SIMPLE      | gameRatingsSandbox0_mv | eq_ref   | PRIMARY       | PRIMARY | 34      | ifdb.games.id | 1    | 1.00   |   100.00 |     100.00 |                                              |
+------+-------------+------------------------+----------+---------------+---------+---------+---------------+------+--------+----------+------------+----------------------------------------------+
```

### DIFF
--- a/sql/patch-schema-2.sql
+++ b/sql/patch-schema-2.sql
@@ -20,6 +20,8 @@ truncate table userScores_mv;
 insert into userScores_mv select *, now() from userScores;
 unlock tables;
 
+alter table games add fulltext key `author` (`author`);
+
 alter table reviews
     add column `embargopastdate` date DEFAULT NULL,
     add key `embargodate` (`embargodate`),
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -670,7 +670,9 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                     // look for this exact name embedded in the author field
                     $expr .= "$or match (author) against ('"
                              . mysql_real_escape_string(quoteSqlLike($n), $db)
-                             . "') ";
+                             . "') and (author like '%"
+                             . mysql_real_escape_string(quoteSqlLike($n), $db)
+                             . "%' ";
 
                     // get the sorting version of the name - LAST, SUFFIX,
                     // FIRST, MIDDLE, and split into an array
@@ -694,6 +696,8 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                                      quoteSqlRLike($nl[0]), $db)
                                      . "[[:>:]]'";
                     }
+
+                    $expr .= ')';
 
                     // join the next round with OR
                     $or = " or";
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -668,9 +668,9 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                 $or = "";
                 foreach ($nameList as $n) {
                     // look for this exact name embedded in the author field
-                    $expr .= "$or author like '%"
+                    $expr .= "$or match (author) against ('"
                              . mysql_real_escape_string(quoteSqlLike($n), $db)
-                             . "%' ";
+                             . "') ";
 
                     // get the sorting version of the name - LAST, SUFFIX,
                     // FIRST, MIDDLE, and split into an array


dfabulich:
In https://github.com/iftechfoundation/ifdb/pull/293#issuecomment-1889399778 I think you were proposing only matching against the last name, instead of the full name. But matching against the full name seems right to me.

Having said that, in testing, I couldn't get any queries to return different results when matching against full vs. last names. (Match always returned a positive, truthy number.)

salty-horse:
> Matching against the full name seems right to me.

My concern (and perhaps misunderstanding of how MATCH works) is that the full name would mean the alternative option of "E. Ryan" won't be selected by MATCH for the LIKE to catch.
If it does return those results, then great!

salty-horse (approved):

dfabulich:
Seems to work on my machine and on dev. I'll give it a day on dev to see if anybody notices trouble.

salty-horse:
Ran some tests locally with initialed and non-initialed names. Seems fine.

salty-horse:
Found A few issues with the match. I think these are pretty minor and can be fixed separately, if you want to push this to the main website soon.

* Discards numbers. So a search for `author:80` won't yield [End of Life Care by 80% Relatable](https://ifdb.org/viewgame?id=695ny3uzvvqeyhfe).
* Doesn't deal with dashes. So `author:b-r-i-a-n` doesn't find [One Angry Poem by b-r-i-a-n](https://ifdb.org/viewgame?id=46a668zq733786ay).
* Likewise, can't click the author name in [Murder by :3](https://ifdb.org/viewgame?id=fl2qc14akizazsge)
* Or [. by .](https://ifdb.org/viewgame?id=6txuyu3ynjwkj3ij), but this doesn't work with the current search either.

I think this is a bug because a like-term-escaping function is used instead of match-term-escaping one (which I used in my initial commit):
* Can't search author with underscores in name. Try clicking the author name in this game: [Let's Get Better or At Least Different by 3_2_1_nothanks](https://ifdb.org/viewgame?id=xpfthekijpnx3uol)
* If an author name is a single word with "%", I think it will also not work for the same reason.

dfabulich:
I found a game where author search used to work in 2023, but now fails, both in production ifdb.org and in this branch. It's "Pizza Quest I" by "C.K." https://ifdb.org/viewgame?id=bifcfkkxxo5034ta 

I tried adding a clause like this, which would ensure that exact-match author searches always work:

```
author = 'C.K.' or ( match (author) against ('C.K.') and (author like '%C.K.%' ))
```

It "worked", but it scanned the full table. I tried adding a non-fulltext index `exact_author` on the author column, but it didn't improve the analyze results. It considered using the indexes, but just gave up and did a full scan.

```
+------+-------------+------------------------+--------+---------------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
| id   | select_type | table                  | type   | possible_keys       | key     | key_len | ref           | rows  | r_rows   | filtered | r_filtered | Extra                                        |
+------+-------------+------------------------+--------+---------------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
|    1 | SIMPLE      | games                  | ALL    | exact_author,author | NULL    | NULL    | NULL          | 12987 | 12987.00 |   100.00 |       0.01 | Using where; Using temporary; Using filesort |
|    1 | SIMPLE      | gameRatingsSandbox0_mv | eq_ref | PRIMARY             | PRIMARY | 34      | ifdb.games.id | 1     | 0.00     |   100.00 |     100.00 |                                              |
+------+-------------+------------------------+--------+---------------------+---------+---------+---------------+-------+----------+----------+------------+----------------------------------------------+
```

I think author search is just too complicated (it's getting pulled in too many directions) to benefit from an index lookup.

* We always want exact-match searches to work, even if the author's name is nothing but punctuation or stopwords
* We also want exact-match _substring_ matches to work (and MySQL/MariaDB has no such index for that)
* When searching by multiple terms, e.g. `author:Firstname Lastname`, we want to match "F* Lastname"
* When the user enters a typo, it'd be nice to get some fuzzy matches, too

This makes me want to reopen and merge #293 after all, to just revert the whole query back to what it was in 2023.

The games table is only 13K rows. MySQL is almost certainly keeping the whole thing in memory pretty much at all times. We've never seen _any_ search show up in the slow query log since we rolled out the materialized views and turned off user_proximity and automated cross-recommendations, which were almost certainly the worst queries, slowing down the entire rest of the site.

Let's make author searches fast someday in the far, far future, when we think it's necessary to do so.

salty-horse:
The B.M. query is probably [too short](https://mariadb.com/kb/en/full-text-index-overview/):
> Words less than 4 (MyISAM) or 3 (InnoDB) characters in length will not be stored in the fulltext index. This value can be adjusted by changing the [ft_min_word_length](https://mariadb.com/kb/en/server-system-variables/#ft_min_word_len) system variable (or, for [InnoDB](https://mariadb.com/kb/en/innodb/), [innodb_ft_min_token_size](https://mariadb.com/kb/en/innodb-system-variables/#innodb_ft_min_token_size)).

If the author query is just digits, letters, and spaces, and >= 4 characters, I think it's safe to use this approach.
Perhaps guard the MATCH term with it? It's a shame to slow down the vast majority of queries when an exception can be made.

If you want, I can try that, and then run a test for all author names currently in the DB, to see if it returns all of the games that just the LIKE query will return.

dfabulich:
I'd like to address https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/436 sooner than later, so I'm going to merge and deploy #293, and leave this PR open. We can do whatever exploration we want to try to fix the fulltext author index after that.

dfabulich:
We don't need this for performance, and author searches are behaving the way they used to thanks to #293, so I'm just going to close this out.

PR #299, opened 2024-02-20
Send TUID to IF Archive

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/449

Now, if possible, we'll add a `tuid` parameter to the request to https://upload.ifarchive.org/cgi-bin/upload.py

There are some catches.

1. If you navigate directly to https://ifdb.org/ifarchive-upload then we won't send a `tuid` parameter. We will send an `ifdbid` "temp ID" parameter. (But it's almost certainly going to be useless, because it won't be associated with a pending link.)

2. Similarly, if you're creating a new game at <https://ifdb.org/editgame?id=new>, as opposed to editing a game, we won't send a TUID in that case, either. Hopefully the user will save the game listing, including the `ifarchive-pending` link, and the "Notify IFDB" button will work anyway.

### DIFF
--- a/www/editgame
+++ b/www/editgame
@@ -1480,17 +1480,15 @@ foreach ($linkfmts as $lf) {
 
 function uploadWindow()
 {
-    var title = document.getElementById("title").value;
-    var devsys = document.getElementById("system").value;
-    var license = document.getElementById("license").value;
-    var author = document.getElementById("eAuthor").value;
-    var version = document.getElementById("version").value;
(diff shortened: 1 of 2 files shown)


erkyrath (approved):
Looks good, not that I know PHP.

salty-horse (approved):

dfabulich:
This is now live.

PR #300, opened 2024-02-21
ifarchive-commit: Create games_history entry when updating url

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/397

I also ported the code to use the `mysqli_execute_query` polyfill (which exposed a bug in the polyfill, which I fixed).

I tested this with TUID 4qfjt6y5pkan2220 (Crystal and Stone, Beetle and Bone).

It has a pending IF Archive link `csbbr4.z8` pointing to `https://ifdb.org/ifarchive-pending?ifdbid=451vzcr3se2vmapo`

So, I was able to locally test it like this:

```
curl "http://localhost:8080/ifarchive-commit?ifdbid=451vzcr3se2vmapo&path=/if-archive/games/zcode/csbb.z8&key="
```

Note that the IF Archive sends us three parameters:

* `ifdbid`: This is the ID of the pending link, _not_ the game's TUID
* `path`: The actual path of the file on the IF Archive
* `key`: A secret key. Using a blank key works for local testing, because the `local-credentials.php.template` uses a blank string for `localIfArchiveKey()`.

See [How IFDB's IF Archive Uploader Works](https://docs.google.com/document/d/1EY4Hs6bPjKg_U-nczs0oP1DqTKv4UPkKnLr0NZ4FqZw/edit) for more details.

### DIFF
--- a/www/ifarchive-commit
+++ b/www/ifarchive-commit
@@ -7,10 +7,13 @@ include_once "dbconnect.php";
 try {
     $db = dbConnect();
 
-    $id = get_req_data("ifdbid");
+    $ifdbid = get_req_data("ifdbid");
     $newPath = get_req_data("path");
-    $oldUrl = "https://ifdb.org/ifarchive-pending?ifdbid=$id";
     $apiKey = get_req_data("key");
+
+    $oldUrl = "https://ifdb.org/ifarchive-pending?ifdbid=$ifdbid";
+
+    $ifarchive_userid = "ix4vb4m6daiyrayh"; // IF Archive Team
 
     // remove any leading '/' in the path
     $newPath = preg_replace("/^\/+/", "", $newPath);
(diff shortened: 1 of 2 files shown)


www/ifarchive-commit
```diff
@@ -4,39 +4,95 @@ include_once "util.php";
...
-$db = dbConnect();
+    $ifdbid = get_req_data("ifdbid");
+    $newPath = get_req_data("path");
+    $apiKey = get_req_data("key");
 
-$id = mysql_real_escape_string(get_req_data("ifdbid"), $db);
-$newPath = mysql_real_escape_string(get_req_data("path"), $db);
-$oldUrl = "https://ifdb.org/ifarchive-pending?ifdbid=$id";
-$apiKey = get_req_data("key");
+    $oldUrl = "https://ifdb.org/ifarchive-pending?ifdbid=$ifdbid";
```
salty-horse, line 14: does `$ifdbid` need to be url-quoted? I assume it's from a trusted source and guaranteed to be safe.

dfabulich: It is from a trusted source, and we verify that by ensuring that the API key is correct.

PR #301, opened 2024-02-21
Add a report of IF Archive pending links

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/399

### DIFF
--- a/www/pending-report
+++ b/www/pending-report
@@ -1,0 +1,84 @@
+<?php
+
+include_once "util.php";
+include_once "pagetpl.php";
+include_once "dbconnect.php";
+
+$db = dbConnect();
+
+$result = mysqli_execute_query($db, "select gameid, title, url from gamelinks where url like '%ifarchive-pending%'");
+
+/*
+there since version N, still there: there's a latest links entry for pagevsn N-1 which does NOT contain the link
+there since version 1: there's not a links entry that does NOT contain the link
+    either because there are no links entries
+    or because all of the links entries have it
+*/
+
+$results = [];
+
+while ($row = mysqli_fetch_row($result)) {
+    [$gameid, $file, $url] = $row;
+    [$title] = mysqli_fetch_row(mysqli_execute_query($db,
+        "select title from games where id = ?", [$gameid]
+    ));
+    [$moddate, $pagevsn] = mysqli_fetch_row(mysqli_execute_query($db,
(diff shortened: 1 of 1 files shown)


salty-horse:
The linked issue suggests this is a feature for moderators. Should the page check your credentials at the top?

www/pending-report
```diff
@@ -0,0 +1,84 @@
...
+there since version N, still there: there's a latest links entry for pagevsn N-1 which does NOT contain the link
+there since version 1: there's not a links entry that does NOT contain the link
+    either because there are no links entries
+    or because all of the links entries have it
+*/
+
+$results = [];
+
+while ($row = mysqli_fetch_row($result)) {
+    list($gameid, $file, $url) = $row;
```
salty-horse, line 21: All the `list(...)` stuff can be replaced with `[...]` - I think it's more readable.

www/pending-report
```diff
@@ -0,0 +1,84 @@
...
+    list($title) = mysqli_fetch_row(mysqli_execute_query($db,
+        "select title from games where id = ?", [$gameid]
+    ));
+    list($moddate, $pagevsn) = mysqli_fetch_row(mysqli_execute_query($db,
+        "select moddate, pagevsn from games_history
+        where id = ? and deltas like '%links%'
+        and deltas not like ?
+        order by pagevsn desc limit 1",
+        [$gameid, "%$url%"]
```
salty-horse, line 30: What do the URLs look like? If they can have `%` and `_`, it might mess with the LIKE.

dfabulich: ifarchive-pending URLs are always supposed to look like this:

```
https://ifdb.org/ifarchive-pending?ifdbid=$ifdbid
```

… where `$ifdbid` is a TUID-like string, consisting only 16 characters, only lowercase letters and numbers.

I guess it might be possible in principle for someone to concoct a link like `https://example.com/ifarchive-pending%20` and that would presumably not work in the `pending-report`, but I don't think that's likely to ever happen, and maybe I'd bother to cross that bridge when I come to it.

dfabulich:
> The linked issue suggests this is a feature for moderators. Should the page check your credentials at the top?

The information shown in the `pending-report` is all public information; anyone who wants to clean up IFDB data could use it. Therefore, I don't see a need to secure it.

PR #302, opened 2024-02-26
Add frame-ancestors CSP rule to prevent clickjacking

dfabulich:
We received a report via email that ifdb.org can be accessed as an iframe on third-party sites, and that this could make IFDB vulnerable to a clickjacking attack.

https://owasp.org/www-community/attacks/Clickjacking

No clear demonstration of a harmful clickjack was demonstrated. I think it would actually be rather challenging to pull off a real clickjacking attack, since our login cookies use `SameSite=Lax`.

But since AFAIK we don't intend anyone to put an IFDB page in an iframe, so we should probably just block iframing; adding the extra protection probably won't hurt anything, and it might help.

### DIFF
--- a/www/csp-nonce.php
+++ b/www/csp-nonce.php
@@ -34,4 +34,5 @@ header("Content-Security-Policy: "
     ."default-src 'self' ifdb.org www.google.com 'nonce-$nonce'; "
     ."script-src 'self' ifdb.org www.google.com 'nonce-$nonce'; "
     ."style-src 'self' ifdb.org 'nonce-$nonce'; "
+    ."frame-ancestors 'self'; "
 );
--- a/www/csp-nonce.php
+++ b/www/csp-nonce.php
@@ -30,4 +30,8 @@ global $nonce;
 $nonce = random_str(8);
 
 // default-src should be enough, but Firefox ESR 102 doesn't support `default-src nonce`, only `style-src nonce` and `script-src nonce`.
-header("Content-Security-Policy: default-src 'self' ifdb.org www.google.com 'nonce-$nonce'; script-src 'self' ifdb.org www.google.com 'nonce-$nonce'; style-src 'self' ifdb.org 'nonce-$nonce';");
+header("Content-Security-Policy: "
+    ."default-src 'self' ifdb.org www.google.com 'nonce-$nonce'; "
+    ."script-src 'self' ifdb.org www.google.com 'nonce-$nonce'; "
+    ."style-src 'self' ifdb.org 'nonce-$nonce'; "
+);


salty-horse:
This is not an area I'm familiar with :)

dfabulich:
The link I pasted explains:

> For example, imagine an attacker who builds a web site that has a button on it that says “click here for a free iPod”. However, on top of that web page, the attacker has loaded an iframe with your mail account, and lined up exactly the “delete all messages” button directly on top of the “free iPod” button. The victim tries to click on the “free iPod” button but instead actually clicked on the invisible “delete all messages” button. In essence, the attacker has “hijacked” the user’s click, hence the name “Clickjacking”.

[...]

> There are three main ways to prevent clickjacking:
>
> 1. Sending the proper Content Security Policy (CSP) frame-ancestors directive response headers that instruct the browser to not allow framing from other domains. The older `X-Frame-Options` HTTP headers is used for graceful degradation and older browser compatibility.
> 2. Properly setting authentication cookies with `SameSite=Strict` (or `Lax`), unless they explicitly need `None` (which is rare).
> 3. Employing defensive code in the UI to ensure that the current frame is the most top level window.

This PR is implementing Option 1.

We already do Option 2, in `session-start.php` and `login-check.php`, explicitly setting `SameSite=Lax`. (Actually, I think all browsers use `SameSite=Lax` by default now, but I guess it's better to be explicit.) `SameSite=Lax` is why I think we aren't _really_ vulnerable to clickjacking.

The bug reporter gave this example code, which he claimed (wrongly, I think) would prove that we're vulnerable to clickjacking:

```
<!DOCTYPE HTML>
<html lang="en-US">

<head>
    <meta charset="UTF-8">
    <title>i Frame</title>
</head>

<body>
    <h3>This is clickjacking vulnerable</h3>
    <iframe src="https://ifdb.org/login?dest=home" height="700px" width="900px"></iframe>
</body>

</html>
```

I claim that this example code doesn't actually prove that we're _vulnerable_, because he's not, for example, lining up the iframe so that clicking on a "win a free iPhone" button automatically rates a game 5 stars or something. _That_ should already be impossible, thanks to `SameSite=Lax`. All he's proving is that it's possible to display IFDB in an iframe.

But, if you pull down this PR's branch and try the example again with `<iframe src="http://localhost:8080/login?dest=home" height="700px" width="900px"></iframe>`, you'll find that the frame will fail to load, because this PR forbids iframing our site.

salty-horse:
I trust this analysis. It seems accurate, but I'm not an expert opinion. In any case, doubt this extra security measure is harmful. If anyone uses this feature and it breaks something for them, they can reach out.

dfabulich:
Deployed

PR #303, opened 2024-02-26
Add a maximum password length to prevent DoS attacks

dfabulich:
We received a report via email that we're vulnerable to a long-password Denial of Service attack.

https://www.acunetix.com/vulnerabilities/web/long-password-denial-of-service/

The idea is: you can try to login with a _huge_ password, forcing us to waste our CPU trying to hash that password. (In a long-password DoS attack, you don't even have to successfully submit the long password; it wastes our memory and CPU just to hash a long password, to check if it matches the hash we keep in the database.)

The reporter was able to submit a password 84 kilobytes long, which honestly probably isn't that bad, especially considering that we still need to fix this https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/11 but it doesn't hurt to address this now.

Note that if there are any users who currently legitimately have a password longer than 1024 characters, their password will be "falsely" rejected, but they'll be able to fix the problem by using the "Lost Password" page to reset their password.

### DIFF
--- a/www/login-check.php
+++ b/www/login-check.php
@@ -96,6 +96,11 @@ function redirect_to_login_page()
 //
 function doLogin($db, $username, $password)
 {
+    if (strlen($password) > MAX_PASSWORD_LENGTH) {
+        return array(false, "incorrect username or password",
+            "The password is too long. If this is your password, please reset it by using the \"Lost Password\" page.");
+    }
+
     // set up a quoted version of the user ID
     $quid = mysql_real_escape_string($username, $db);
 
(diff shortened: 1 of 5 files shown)


www/login-check.php
```diff
@@ -96,6 +96,12 @@ function redirect_to_login_page()
 //
 function doLogin($db, $username, $password)
 {
+    if (strlen($password) > MAX_PASSWORD_LENGTH) {
+        return array(false, "incorrect username or password",
```
salty-horse, line 100: Is there a security reason to not explain the actual error?

> The password is too long. If this is your password, please reset it by using the "Lost Password" page

salty-horse (approved):

dfabulich:
Deployed

PR #306, opened 2024-07-26
Add IFTF donation badge on home page and footer

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/167

# Home page

image

# Footer

image

# Mobile home page

# Mobile footer

# Mobile dark mode


erkyrath:
You've got the IF Archive ID in there... should be `2CZJEA3PXM3QS` instead.

dfabulich:
Oops! Fixed now.

salty-horse:
Shouldn't the button be [branded](https://developer.paypal.com/docs/multiparty/checkout/standard/customize/buttons-style-guide/) with the PayPal logo and the yellow color?

I understand if you want to keep the page simple. But people might be more psychologically comfortable pressing a recognizable button.

The branding guide above is for *payment* buttons. I don't have access to the actual SDK config without logging in. Do they recommend other visuals for *donation* buttons? I can only find this [old example](https://developer.paypal.com/api/nvp-soap/paypal-payments-standard/integration-guide/html-example-donate/), which uses an image for the button. (I'm not recommending using it - just as an example of what it could look like.)

erkyrath:
We've drifted away from using the Paypal branding on our sites. It's recognizable, but it's an image button (rather than a native HTML button) and the color clashes everywhere. Using a simple button with CSS that blends with the site seems to work fine.

https://iftechfoundation.org/give/
https://ifarchive.org
https://narrascope.org/pages/sponsors.html

I mean, we know people push it!

salty-horse:
Those other buttons at least have some color to make them pop :)

dfabulich:
I've added color and style to the donation button and updated the screenshots.

www/ifdb.css
```diff
@@ -1066,6 +1066,35 @@ div.footer {
...
+    float: left;
+    height: 6em;
+    padding-right: 2ch;
+}
+
+.iftf-donation form {
+    padding-top: 1em;
+    margin: 0;
+}
+.iftf-donation .Donate {
```
salty-horse, line 1089: why upper-case?

salty-horse (approved):

www/ifdb.css
dfabulich: Copied and pasted from IF Archive. 🤷‍♂️

PR #308, opened 2024-07-27
Add a link to TADS.org on the home page side bar.

dfabulich:
Apropos https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/235

image


erkyrath (approved):
Yep, looks good.

PR #309, opened 2024-07-29
Add hover styling to donation button

dfabulich:
@erkyrath wrote:

>  I’d recommend adding `cursor: pointer;` to the CSS for the button. The link feels unresponsive without that.

… I didn't actually do that, in this case, I just added a darker background to the hover state, but I'm open to discussing it.

I didn't add `cursor: pointer` because this a button, not a link. When you click it, it issues a POST action to PayPal, which you're not supposed to refresh.

By default, HTML buttons don't have `cursor: pointer`. When you hover them, they get a little darker. `cursor: pointer` was designed for underlined links, to make it clear to users that the text is _clickable_. 

https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b
> Buttons shouldn't have a hand cursor.

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -1093,6 +1093,11 @@ div.footer {
     padding: 8px 10px;
     color: white;
     font-weight: bold;
+    cursor: pointer;
+}
+
+.iftf-donation .Donate:hover {
+    background-color: #b45620;
 }
 
 /*


erkyrath:
It is true that the default stylesheets for `<button>` and `<input type="submit">` don't have the hand cursor. However, I feel like web design has moved decisively to the hand cursor for buttons that go somewhere or submit a form. (Quick survey including google.com, amazon.com, developer.mozilla.org, and this github page all consistently agreed.) That blog post is making an argument from consistency, but surely you can make the opposite argument *also* from consistency.

I note that the default stylesheet for `<input type="image">` *does* have a hand cursor, and a lot of Paypal buttons use image buttons. They look exactly like the text-type buttons except for color and font.

All that said, I'm not going to argue further if you want to leave it as-is. 

> I just added a darker background to the hover state

I'm all in favor of that, as you can see from the other IFTF pages.

dfabulich:
eh, you've talked me into it. Now it includes `cursor: pointer`.

I'll merge it after somebody approves the PR

PR #310, opened 2024-07-29
Remove donation ad from IFDB home page

dfabulich:
Not sure if we want to do this, but folks are complaining that the donation ad is too prominent on the home page, and this is the easiest way to fix that.

(The donation ad would still appear in the footer on all pages, including the home page.)


salty-horse:
How about above the sidebar, with a smaller logo? Still prominent, but not center stage.

Might not look good on mobile, though.

salty-horse (approved):

dfabulich:
I think we're going with the collapsible Welcome page instead.

PR #311, opened 2024-07-30
Collapsible welcome section on home page

dfabulich:
image

image

### DIFF
--- a/www/home
+++ b/www/home
@@ -36,12 +36,15 @@ if ($debugflag) echo "debug mode enabled...<br>";
    <div class="column col-main">
 
       <div class="block">
-         <div class="headline headline1">Welcome to IFDB!
-            <span class=headlineRss>
-               <a href="news">Site News</a> |
-               <a href="news?rss"><img src="img/blank.gif" class="rss-icon" alt="">RSS Feed</a>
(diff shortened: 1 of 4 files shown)


www/home
```diff
@@ -36,12 +38,34 @@ if ($debugflag) echo "debug mode enabled...<br>";
    <div class="column col-main">
 
       <div class="block">
-         <div class="headline headline1">Welcome to IFDB!
-            <span class=headlineRss>
-               <a href="news">Site News</a> |
-               <a href="news?rss"><img src="img/blank.gif" class="rss-icon" alt="">RSS Feed</a>
-            </span>
-         </div>
+         <details id="welcome-details" <?= $welcomeOpen ? "open" : ""?>>
```
salty-horse, line 41: Since this is clickable, it needs an ARIA role. I think should be a [button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role), but I'm not an expert.

dfabulich: My understanding is that the `<details>` element is already automatically ARIA enabled. (That's what makes the clickability work.)

salty-horse (approved):

PR #320, opened 2024-08-12
Fix poll news on home page

dfabulich:
The home page is supposed to list new polls in the "New on IFDB" section, but the query always failed, because the `users` table is aliased as `u`.


salty-horse:
I fixed this as part of #260 - had no idea it was causing problems!

PR #323, opened 2024-08-18
Fix IF Archive scrubber: Scrub troll content, but don't delete the user

dfabulich:
The IF Archive scrubber was deleting sandboxed (troll) users, and half-heartedly deleting their reviews and their comments. But, looking at https://github.com/iftechfoundation/ifdb/blob/main/www/util.php#L2785 the `close_user_acct` had a _bunch_ more stuff to do that the scrubber wasn't considering.

For example, poll votes from sandboxed users were included in the archived database, even though the sandboxed user rows were deleted.

Here, I copied a bunch of stuff from `close_user_acct` into the scrubber, and I changed it to _not_ delete the user at the end. Instead, with all of sandbox users' comments, reviews, votes, etc. etc. deleted, it appears as if the user has had little/no site activity.

… but one thing that _is_ preserved is the user's edit history. Previously, when viewing the edit history of a game in local development, sandboxed users' edits would appear with a blank name. Now, they are correctly linked to the sandboxed user who made the edits.

### DIFF
--- a/sql/ifdb-to-ifarchive-scrubber.sql
+++ b/sql/ifdb-to-ifarchive-scrubber.sql
@@ -50,13 +50,46 @@ drop trigger reviews_insert;
 drop trigger reviews_update;
 drop trigger reviews_delete;
 
-/* delete troll users and their reviews and comments */
+/* delete troll reviews and comments */
 delete from ucomments
-  where userid in (select id from users where sandbox = 1);
-delete from reviews
-  where userid in (select id from users where sandbox = 1);
-delete from users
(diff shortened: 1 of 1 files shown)


Keltena (commented):
Looks like an improvement from what I can see!

I noticed that the scrubber doesn't touch gametags by sandboxed users. Is that how we want it, at least until iftechfoundation/ifdb-suggestion-tracker#423 is addressed?

dfabulich:
I think so, and I think we should update the scrubber if/when we fix that bug.

PR #326, opened 2024-08-20
Hide sandboxed explicit cross-recommendations

dfabulich:
This one's a bit tricky to test, because the UI automatically hides your own cross-recommendations (and anything you've played, wishlisted, or unwishlisted).

So, to see the effect, you first have to create two sandboxed users, each of whom makes an explicit cross-rec to a game with no cross-recs. Each one will see the other's cross-recs. Before the PR, logged out users would see both cross-recs, but after this PR, they see neither.

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -19,8 +19,8 @@ include_once "news.php";
 $userprivs = $adminPriv = false;
 if ($curuser) {
     $result = mysqli_execute_query($db,
-        "select `privileges` from users where id=?", [$curuser]);
-    $userprivs = mysql_result($result, 0, "privileges");
+        "select `privileges`, `sandbox` from users where id=?", [$curuser]);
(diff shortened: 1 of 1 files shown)


www/viewgame
```diff
@@ -1676,6 +1670,7 @@ if (!isEmpty($genre)) {
         }
 
         // check for explicit cross-recommendations
+        $sandbox = $mysandbox ? "1" : "and not(u.sandbox) and not (u2.sandbox)";
```
salty-horse, line 1673: The "else" string starts with `and`, but the SQL below already does `and $sandbox`.

Doesn't that expand to `and and`?

dfabulich: Good catch! Fixed.

salty-horse (approved):

PR #332, opened 2024-09-18
Fix float-clearing bug

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/491

In #331, I removed what I _thought_ was an "unnecessary" table, with just one row and one cell.

But the table _did_ serve a purpose that I didn't realize at the time: it created a [block formatting context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flow_layout/Introduction_to_formatting_contexts).

When the `externalLinksDiv` went to `clear: both`, when the table was there, it would only clear the `coverart`, but when we removed the table, it would also attempt to clear the External Links div, which could be arbitrarily long, creating a huge empty space on the page.

Adding `display: flow-root` to the `mainSummary` fixes the issue.

I've tested this on Counterfeit Monkey, Anchorhead, Uninteractive Fiction, and [Cold Iron](https://ifdb.org/viewgame?id=x8ohk12d1a6f12ge).

I missed this in #331 because I'd only tested Uninteractive Fiction and Counterfeit Monkey, but CM doesn't have a Website and so it didn't have an `externalLinksDiv` on desktop screen sizes, so it didn't trigger this bug. Anchorhead does, and so does Cold Iron (which is the case where the bug was first reported to me).

This bug is so bad that I think I'm going to merge the fix without review, but I invite you to take a look after the fact!

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -873,7 +873,7 @@ echo helpWinLink("help-ifid", "IFID");
 
 ?>
 <style nonce="<?php global $nonce; echo $nonce; ?>">
-    .viewgame__mainSummary { margin: 0; padding: 0; order: -1; }
+    .viewgame__mainSummary { margin: 0; padding: 0; order: -1; display: flow-root; }
     @media (max-width: 60rem) {
         #viewgame-body {
             display: flex;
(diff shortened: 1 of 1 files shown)


erkyrath:
Looks right at a quick glance. (`display: flow-root` is one of my favorite recent discoveries!)

PR #334, opened 2024-09-24
search: Make `played:yes` filter include games you've reviewed/rated

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/494

### DIFF
--- a/www/search
+++ b/www/search
@@ -972,7 +972,8 @@ else {  // ...default is searching games
             the given id.
 
        <p><b>played:<i>yes</i>|<i>no</i></b> lists games that
-            you have/haven't put on your played list (i.e. "I've played it").
+            you have/haven't put on your played list (i.e. "I've played it")
+            and have/haven't reviewed or rated.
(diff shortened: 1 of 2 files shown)


salty-horse:
I wonder if it should work differently, where rating a game will also mark it as "played", with a flag that it was done automatically due to a rating.

And delete all ratings/reviews for a game, it should check if "played" was an automatic marking, and delete that too.

salty-horse (approved):

salty-horse:
Does it make sense to have a separate `reviewed:yes` query to only search games you're reviewed?

dfabulich:
I'm unsure what the right approach is; I opened a thread about it on the intfiction forum here. https://intfiction.org/t/ifdb-played-yes-and-ratings-reviews/71269

dfabulich:
I'm closing this out, in favor of https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/496

PR #336, opened 2024-09-24
Replace image buttons

dfabulich:
Replace old GIF-based image buttons with HTML buttons.


www/editclub
```diff
@@ -478,16 +478,15 @@ formCol("membersPublic", $membersPublic, "Membership",
...
 
-echo "<tr><td></td><td><p><input type=image alt='Save Changes' "
-   . "value='Save Changes' name='save' "
-   . "src='img/blank.gif' class='save-button'> &nbsp; "
+echo "<tr><td></td><td><p><button value='Save Changes' name='save' "
+   . "class='fancy-button'>Save Changes</button> &nbsp; "
 
-   . "<input type=image alt='Reset Form (Discard Changes) "
+   . "<button class='fancy-button' "
    . "value='Reset Form (Discard Changes)' name='discard' "
```
salty-horse, line 485: Is this `value='Reset Form (Discard Changes)'` used by anything? Can it be removed? (Also relevant to other changes in this MR)

I thought it's only relevant when submitting form data, but `name` is already used to identify the button. Is it exposed to screen readers?

salty-horse (commented):
There are some comments still mentioning "image button" that should be changed.

www/editclub
dfabulich: I didn't know, and since I was making a bunch of changes, I thought it might be safer to just leave it in. Seems harmless?

salty-horse (approved):

dfabulich:
> There are some comments still mentioning "image button" that should be changed.

I yanked out a bunch of spots where image button `_x` parameters were being translated. They're no longer needed.

PR #337, opened 2024-09-25
Add `reviewed:yes` and `rated:yes` filters

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/495

Apropos #334, it's not totally clear that #334 is the right approach, and it's not clear that we'll get to a consensus quickly. But adding search filters is uncontroversial, and this alone allows users to get the search I _actually_ care about, which is `played:no willplay:no wontplay:no reviewed:no rated:no`, a search filter for games that are "new to you."

If we decide to merge #334, then `played:no` will effectively mean `played:no reviewed:no rated:no`. Alternately, if we automagically mark games as "I've played it" when you review or rated it, `played:no` will, similarly, effectively mean `played:no reviewed:no rated:no` because everything you've rated or reviewed will be added to your "I've played it" list.

But it also means that you'll be able to use `played:yes reviewed:no rated:no` to find games that you've played but not reviewed or rated, which could be useful in its own right.

### DIFF
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -254,6 +254,8 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
             "played:" => array("played", 99),
             "willplay:" => array("willplay", 99),
             "wontplay:" => array("wontplay", 99),
+            "reviewed:" => array("reviewed", 99),
+            "rated:" => array("rated", 99),
             "license:" => array("license", 0),
             "format:" => array("/gameformat/", 99));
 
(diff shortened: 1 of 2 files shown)


www/searchutil.php
```diff
@@ -659,6 +661,42 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                 }
                 break;
 
+            case 'reviewed':
+                // Only use this query when the user is logged in
+                if ($curuser) {
```
salty-horse, line 666: I know the rest of the code is like this, but I prefer using "guard"-style checks to avoid extra indentation.
```php
if (!$curuser) {
    break;
}

// rest of code
```

salty-horse (approved):

PR #340, opened 2024-09-27
Search by competitionid

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/387

### DIFF
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -604,6 +604,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                 // need to join the gameprofilelinks table to do this query
                 if (!isset($extraJoins[$col])) {
                     $extraJoins[$col] = true;
+                    $txt = mysql_real_escape_string($txt, $db);
                     $tableList .= " inner join gameprofilelinks as gpl "
                                   . "on gpl.gameid = games.id "
                                   . "and gpl.userid = '$txt'";
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -257,6 +257,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
             "reviewed:" => array("reviewed", 99),
             "rated:" => array("rated", 99),
             "license:" => array("license", 0),
+            "competitionid:" => array("competitionid", 99),
             "format:" => array("/gameformat/", 99));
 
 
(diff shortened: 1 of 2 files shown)


www/searchutil.php
```diff
@@ -609,6 +610,16 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
...
                 }
                 break;
+            
+            case 'competitionid':
+                // need to join the gameprofilelinks table to do this query
+                if (!isset($extraJoins[$col])) {
+                    $extraJoins[$col] = true;
+                    $tableList .= " inner join compgames "
+                                  . "on compgames.gameid = games.id "
+                                  . "and compgames.compid = '$txt'";
```
salty-horse, line 620: Does `$txt` need to be escaped with `$txt = mysql_real_escape_string($txt, $db);`?

It's done in several other places, but not in `authorid` and here.

Is it guaranteed to be safe as part of an earlier parsing step? I can't find it.

www/searchutil.php
```diff
@@ -609,6 +610,16 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                                   . "and gpl.userid = '$txt'";
                 }
                 break;
+            
+            case 'competitionid':
+                // need to join the gameprofilelinks table to do this query
```
salty-horse, line 615: wrong table name in copy-pasted comment.

www/viewcomp
```diff
@@ -407,7 +407,8 @@ newsSummary($db, "C", $compID, 3,
             "</div>");
 
 // show the games, by division
-echo "<h2>Games and Awards</h2><div class=indented>";
+echo "<style nonce='$nonce'>.compgames { display: inline-block; } </style><h2 class=compgames>Games and Awards</h2>
+    <span class='details'><a href='/search?searchbar=competitionid:$compID'>Search for games</a></span> <div class=indented>";
```
salty-horse, line 411: I don't like the label "Search for games". It's a "List all participating games" that happens to use the site's search functionality.

More accurately, it's a list of all the participating games that are on IFDB - maybe that should be clarified somewhere?

A bigger question: Perhaps the competition page needs a proper section with the list of all games (and if it's too long, lead to the search page)

www/searchutil.php
dfabulich: Nope! It's just 100% vulnerable. 😬

www/viewcomp
dfabulich: > A bigger question: Perhaps the competition page needs a proper section with the list of all games (and if it's too long, lead to the search page)

Not sure what you mean by this. The existing "Games and Awards" does show all of the competition's entrants already. But it shows the games in a hardcoded order, without cover art or ratings info, and there's no way to sort or filter the competition results.

You want to see a list of IFComp 2024 games sorted by Highest Rating? You want to exclude Twine games? You want to search for that IFComp game about "aliens" but you can't remember the name? Today, the only way to do any of that is to click through on a random game in IFComp 2024, then scroll down and click on the "IFComp 2024" tag, and then filter your search within that.

By linking to a search for `competitionid:4lxgmwam7owmbb9h`, we make it more obvious how to do all the rest of the search stuff.

In light of that, is "Search for games" still not the right answer here? Is there something better we can think of? I think it's more than just "List all participating games," especially since the list is already right there…?

dfabulich: I've moved the link out to a separate PR #341.

salty-horse (approved):

PR #342, opened 2024-09-29
editcomp: Fix broken table layout when one game has a long name

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/498

Before:
image

After: 
image

### DIFF
--- a/www/combobox.php
+++ b/www/combobox.php
@@ -23,7 +23,7 @@
 function makeComboBox($name, $textlen, $curval, $vals, $onSet = "null")
 {
     // write the input field
-    $txt = "<input type=\"text\" name=\"$name\" id=\"$name\" "
+    $txt = "<span class='nobr'><input type=\"text\" name=\"$name\" id=\"$name\" "
            . "size=$textlen value=\"" . htmlspecialcharx($curval)
            . "\">"
            . addSiblingEventListeners([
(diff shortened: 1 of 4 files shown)


salty-horse:
MDN [warns not to use `<nobr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr), which is deprecated, and suggests `style="white-space: nowrap;"` instead.

Will it work for these elemenets?

dfabulich:
Replaced with `<span class='nobr'>` which does the same thing.

www/combobox.php
```diff
@@ -23,7 +23,7 @@
 function makeComboBox($name, $textlen, $curval, $vals, $onSet = "null")
 {
     // write the input field
-    $txt = "<input type=\"text\" name=\"$name\" id=\"$name\" "
+    $txt = "<span class='nobr'><input type=\"text\" name=\"$name\" id=\"$name\" "
```
salty-horse, line 26: Now there's a mix of `"` and `'` in the generated HTML, in case it matters to you.

salty-horse (approved):

www/combobox.php
dfabulich: doesn't matter to me 🤷

PR #343, opened 2024-09-30
putific: allow API to reorder or delete existing links

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/497

### DIFF
--- a/www/putific
+++ b/www/putific
@@ -578,8 +578,18 @@ if ($linkFile)
     // get the original links from the existing record, if any
     $recLinks = (isset($rec["links"]) ? $rec["links"] : array());
 
+    $replaceLinks = get_req_data("replaceLinks") === "yes";
+    if ($replaceLinks) {
+        $reqLinks = [];
+    } else {
+        $reqLinks = $recLinks;
+        $existingUrls = [];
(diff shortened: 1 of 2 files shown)


www/putific
```diff
@@ -603,26 +609,28 @@ if ($linkFile)
...
-        // as a new link.
-        for ($i = 0, $found = false ; $i < count($recLinks) ; $i++)
-        {
-            $recl = $recLinks[$i];
-            if (strcasecmp($recl["url"], $url) == 0)
+        if (!$replaceLinks) {
+            // Look for an existing link with the same URL.  If there's an
+            // existing link, we'll update it.  Otherwise we'll add this one
+            // as a new link.
+            for ($i = 0, $found = false ; $i < count($recLinks) ; $i++)
```
salty-horse, line 616: Instead of this nested loop, I'd populate an array above the top loop with lowercased versions of the original URL pointing at their `rec link` objects.

Then the lookup is just an O(1) check if the lowercased new URL is in the array. (Not that speed matters in this case, but I think it's more readable)

salty-horse (approved):

PR #344, opened 2024-09-30
taggame: Fix login via API

dfabulich:
#325 broke the `taggame` API, which passes a username and password as POST parameters. In that case, we never populated `$acctstatus`, `$profilestatus`, or `$sandbox`, so the code was rejecting all tags via API.

### DIFF
--- a/www/taggame
+++ b/www/taggame
@@ -66,26 +66,26 @@ for ($i = 0, $tags = array() ; ; $i++) {
 
 // make sure we process any persistent login state
 $userid = checkPersistentLogin();
-if ($userid) {
-    $result = mysqli_execute_query($db, "select acctstatus, profilestatus, sandbox from users where id=?", [$userid]);
-    if (!$result || mysql_num_rows($result) == 0) {
-        $userid = null;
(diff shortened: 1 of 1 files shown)


www/taggame
```diff
@@ -66,6 +66,11 @@ for ($i = 0, $tags = array() ; ; $i++) {
 
 // make sure we process any persistent login state
 $userid = checkPersistentLogin();
+if (!$userid && $xml) {
+    $username = get_req_data("username");
+    $password = get_req_data("password");
+    list($userid, $errCode, $errMsg) = doLogin($db, $username, $password);
```
salty-horse, line 72: If you're moving code, might as well change `list(x, y, z) =` to `[x, y, z] =`, which is I find cleaner, and is supported since PHP 7.1.

salty-horse, line 72: I think a login failure should return immediately, before attempting to do anything else.

So move the `if (!$userid) { sendResponse("401 Unauthorized") }` here.

Is there a reason not to do that...?

salty-horse (approved):

PR #345, opened 2024-09-30
editcomp: Add games by tag

dfabulich:
(Thanks for reviewing all of these PRs!)

### DIFF
--- a/www/editcomp
+++ b/www/editcomp
@@ -1330,7 +1330,8 @@ function nthOrd(n)
         echo "var divModel$i = { "
             . "addbutton: 'Add a Game', "
 
-            . "addExtra: \" <a href='#' id='sortDivByPlaceButton$i' class='fancy-button'>Sort by Place</a>\", "
+            . "addExtra: \" <a href='#' id='sortDivByPlaceButton$i' class='fancy-button'>Sort by Place</a> "
+            .   "<button id='addByTag$i' type=button class='fancy-button'>Add Games by Tag</button>\", "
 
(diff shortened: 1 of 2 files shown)


salty-horse (approved):
Didn't run it, but it looks fine. I assume the code is copied from `importers/entrant-creator.js`.

Speaking of the search API, would it be useful to have json output in addition to xml?

www/editcomp
```diff
@@ -1461,6 +1462,50 @@ function activateListeners(divno, model)
...
+        dialog.innerText = "Loading";
+        document.body.append(dialog);
+        dialog.showModal();
+
+        const url = '/search?' + new URLSearchParams({
+            searchbar: 'tag:' + tag,
+            xml: 1,
+            pg: 'all',
+            sortby: 'ttl',
+        }).toString();
```
salty-horse, line 1478: I think `toString()` isn't needed.

And this might be more readable:
```javascript
const params = new URLSearchParams(...);
...
await fetch(`/search?{$params}`);
```

dfabulich:
> Speaking of the search API, would it be useful to have json output in addition to xml?

It absolutely would be useful. The only reason I haven't done it is that it seems like a lot of work, and plausibly a lot of maintenance burden.

salty-horse:
For the assumed maintenance burden, do you mind scanning the web server logs to discover how many consumers it has? If it's not too many, maybe it can be deprecated in a reasonable time frame.

We can move this conversation to the forums.

PR #346, opened 2024-10-02
Ifarchive-tuid-report: normalize urls

dfabulich:
This required me to add composer support, but I struggled with it mightily. I'm not at all sure I'm handling that the "right way."

In the previous draft iteration of this PR, I checked in the entire `vendor` directory, which doesn't seem to be the "right way."

Now, I'm using a Docker multi-stage build in the `Dockerfile` to run `composer install` as the image is built. I've also added a new `docker-php-composer.sh` script that you can run to update `composer.json` and `composer.lock` and install new dependencies.

I think this PR is much cleaner than the previous version; it should be ready to merge.

### DIFF
--- a/www/ifarchive-tuid-report
+++ b/www/ifarchive-tuid-report
@@ -3,6 +3,7 @@
 include_once "util.php";
 include_once "pagetpl.php";
 include_once "dbconnect.php";
+require_once 'vendor/autoload.php';
 
 $json = $_GET['json'] ?? false;
 $refresh = $_GET['refresh'] ?? false;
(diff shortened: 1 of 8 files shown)


www/ifarchive-tuid-report
```diff
@@ -87,9 +88,20 @@ $db = dbConnect();
 
 $output = [];
 foreach($tuids as $tuid => $path) {
-    $result = mysqli_execute_query($db, "select url from gamelinks where gameid = ? and url like ?", [$tuid, "%$path"]);
-    $row = mysqli_fetch_row($result);
-    if (!$row) {
+    // normalize the path using URL normalizer. First we make it an absolute URL, normalize it, then we strip the absolute part off
+    $path = substr((new URL\Normalizer("https://ifarchive.org/$path"))->normalize(), strlen("https://ifarchive.org/"));
```
salty-horse, line 92: I would put `https://ifarchive.org/` in a variable since it's used twice, and explicitly required to be identical for this "get suffix" technique.

salty-horse (commented):
I don't know anything about composer and PHP package management.

Based on other packaging systems, I find it odd that the package is mentioned in `composer.json` but is also checked into the repository.

dfabulich:
I'm going to merge this now, because there's another kinda time-sensitive feature that requires access to third-party dependencies.

PR #348, opened 2024-10-04
Add message about the "Use Accessible Controls" setting

alice-blue:
People who might like to use the "Use accessible controls" setting on the settings page may not know it exists. This message is meant to make users aware of that setting, in text that is read by a screen reader but is not visible on the screen.

I have not been able to test it, but it is modeled after the following code. An IFDB user confirmed that the "Yip yip yip" text in the code below is read aloud by a screen reader, even though it is not visible on the screen:

```
<h1>Woof!</h1>
<span style="position:absolute;width:1px;left:-10000px;">Yip yip yip</span>
```

### DIFF
--- a/www/starctl.php
+++ b/www/starctl.php
@@ -51,6 +51,7 @@ function setStarCtlValue(id, val)
         // with automatic mouse rollover highlighting.
 
         ?>
+<span style="position:absolute;width:1px;left:-10000px;">To rate games using a keyboard, please visit the settings page and turn on the "Use Accessible Controls" setting.</span>
 <script type="text/javascript" nonce="<?php global $nonce; echo $nonce; ?>">
 <!--
 var starRatings = {};
--- a/www/starctl.php
+++ b/www/starctl.php
@@ -51,7 +51,7 @@ function setStarCtlValue(id, val)
         // with automatic mouse rollover highlighting.
 
         ?>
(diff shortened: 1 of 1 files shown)


salty-horse:
It's important to mention that this addresses https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/509.

As mentioned there, the correct fix is to fix the star rating control to not require a special accessibility opt-in, which I'll try to tackle soon. Until that is done, this message is good to have.

I don't like the pixel offset trickery. [This Stack Overflow answer](https://stackoverflow.com/a/26032207) suggests this instead:
```html
<div aria-label="text only visible to screen readers"></div>
```

alice-blue:
Hmm...there are comments on that Stack Overflow answer with some people saying the aria-label suggestion didn't work for them. One commenter says "Microsoft Edge "Read Aloud" ignores aria-label and 1px elements (width or height). Only option 3 works on it." So maybe width:1px isn't a good idea either.

alice-blue:
Ok, now that I've actually tested this, I can see that it doesn't even hide the text at all. Maybe because it's in the "Have you played this game?" box, the position thing doesn't work?

dfabulich:
I'm hoping we can do #350 instead.

dfabulich:
The star control is now accessible by default.

PR #349, opened 2024-10-04
Add visible message about accessibility option

alice-blue:
This is another way to approach https://github.com/iftechfoundation/ifdb/issues/370

This would be a visible message, and would be temporary until the star rating controls are redone.

I am not sure how to test this in the dev environment, because it requires being logged in, and I have not been able to activate a new user in the dev environment in order to log in.

### DIFF
--- a/www/starctl.php
+++ b/www/starctl.php
@@ -51,6 +51,7 @@ function setStarCtlValue(id, val)
         // with automatic mouse rollover highlighting.
 
         ?>
+<span class=fmtnotes>To turn on accessible controls, please visit the <a href="https://www.ifdb.org/editprofile">settings page</a>.</span>
 <script type="text/javascript" nonce="<?php global $nonce; echo $nonce; ?>">
 <!--
 var starRatings = {};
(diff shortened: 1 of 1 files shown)


www/starctl.php
```diff
@@ -51,6 +51,7 @@ function setStarCtlValue(id, val)
         // with automatic mouse rollover highlighting.
 
         ?>
+<span class=fmtnotes>To turn on accessible controls, please visit the <a href="https://www.ifdb.org/editprofile">settings page</a>.</span>
```
salty-horse, line 54: the href should be just `/editprofile`.

salty-horse:
I edited the HTML in my browser to simulate how it looks. Seems fine to me:
ratings_box

www/starctl.php
alice-blue: Thanks! I fixed the href. I also got a chance to log in and test it, and added <br> so that the stars will start on a new line when this message shows up on the "Write a Review" page.

dfabulich:
I'm hoping we can do #350 instead.

dfabulich:
The star control is now accessible by default.

PR #350, opened 2024-10-05
Accessible star controls

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb/issues/490

Based on https://dev.to/grahamthedev/5-star-rating-system-actually-accessible-no-js-no-wai-aria-3idl

Tested on iOS 18 screen reader. It reads display ratings aloud, e.g. "4 and a half stars out of 5", and it supports navigating to each star as a radio button and checking it.

image

image

image

image

image

image

image

image


salty-horse:
LOL I was just working on it. [Here's mine](https://github.com/iftechfoundation/ifdb/compare/main...salty-horse:ifdb:accessible_star_rating?expand=1), still has a few bugs (see TODO).

I'll review your approach with the knowledge I gained :)

salty-horse:
UI issues I noticed:
1. Since the radio buttons have no "confirmation" step, just changing the focus over them with the keyboard changes the selection (and you made it submit). I don't think this is friendly to keyboard users , who then need to scramble to figure out how to undo their action. From an HTTP angle, I don't think data should be POST-ed to a server just by tabbing over elements. I started my own attempt with radio buttons and switched to regular clickable buttons (Goodreads-style) because of this. I believe radio buttons should only be used when you have a form with a separate "submit" button.
2. If you rated something as 3 stars, then want to change it to 5, It's weird that 3 stars are red, and 2 are yellow.
3. After rating something, the "remove it" should show up immediately. (The old code tied it to a mouse-out, which is wrong. That mouse-out is only needed to change the "Rate it:" to "Your Rating", to avoid pushing the stars element to the right as your hovering over it.)
4. Missing ARIA labels: The fieldset should have an aria-label specifying the current rating (or no rating), and updated along with the server. The radio buttons labels should say "1 star" and not just "1".
5. This is an issue with the current implementation that I think should be fixed, and probably can't be fixed with the radio buttons' "submit on focus change": If there was an issue sending the data to the server, I think the UI should revert back to the state before the click.

Cleanup:
1. The review page still uses the accessibility setting. It should be changed.
2. The settings page should say that the accessbility toggle currently does nothing :)

General idea from my own attempt:
1. Move JS code out of individual pages' PHP soup into `ifdbutil.js`, and add `'use strict;'` at the top. It makes it much easier to catch errors, and for linters to do their job.

alice-blue:
> 2\. The settings page should say that the accessbility toggle currently does nothing :)

I wonder if there's anyone who likes the drop-down menu and would want to keep using it. If so, the description could be changed to say "Use drop-down menu for star ratings."

salty-horse:
If the new implementation is good enough, there's no need to keep the drop-down. Can we consult with actual users of this feature, even if it's just to get feedback on the new implementation?

For the setting, it might be good to keep for future use...?

dfabulich:
I started working on this again but didn't really see a clever way to implement this that works equally well on the `viewgame` page and on the `review` page, unless I add a "submit" button to `viewgame`, which barely fits, and doesn't look great.

@salty-horse do you want to take a crack at this?

salty-horse:
@dfabulich I'm not sure what is the "this" you mean that causes trouble. Something in my list above? I can see how switching from radio to buttons means a bit more work with the `review` form.

dfabulich:
"This" is just, this PR, in general.

I began by saying, "OK, I'll turn the radio buttons into buttons" and then I said, "but what about reviews?"

So then I went in to add a "Submit" button on `/viewgame` but it didn't really fit (especially when the stars are large enough to tap with a finger). Then I got tired of working on it and wandered off!

salty-horse:
OK, I'll fix the remaining stuff.

BTW, in Goodreads' "review" form, if you change the star rating, it submits it immediately, separately from the rest of the form. So it might be an OK behavior, but I prefer the current one.

salty-horse:
Pushed commits. Please review.

Outstanding issues in my commits:
1. There's no visual indicator when you hover over a star and click. In `viewgame` there's a "saved" label that appears, but if you click twice you will not know something happens. In `review` and `editgame` (when adding external reviews) it's missing. I think it should turn from yellow to red after clicking.
2. All 3 places that use the star control have a "remove rating", and there's some code duplication about removing that link when there's no rating. I wonder if it's better to have is an "X" to the right of the stars, and make it managed by the star control and the click callback.
3. I couldn't get the button focus highlighting of the stars to work, esp. together with the hover highlighting. Maybe `:focus-visible` doesn't play well with the `+` and `~` selector tricks (see TODO.) I might switch to styling them in javascript, which will give better control about the first issue mentioned above.
4. In `viewgame`, I don't like the label being changed from "Rate it:" to "My Rating:" pushes the "saved" to the next row. Also, should that be "Your Rating:" instead?
5. The aria label for the control says "Your rating is X out of 5", and "You did not rate this game", but in `editgame`, it's not really your rating - you're inputting the ratings of external reviews. Should it be changed to something more appropriate?
6. In `viewgame`, when submitting the rating after a click, if we failed to contact the server, or it returned an error, I think the user should be alerted. Since it's rare, is it OK to use an `alert()`?

alice-blue:
> 5\. The aria label for the control says "Your rating is X out of 5", and "You did not rate this game", but in `editgame`, it's not really your rating - you're inputting the ratings of external reviews. Should it be changed to something more appropriate?

Maybe for external reviews, it could say "You entered the rating X out of 5." and "You did not enter a rating for this review."

salty-horse:
> Maybe for external reviews, it could say "You entered the rating X out of 5." and "You did not enter a rating for this review."

Or it could be changed everywhere to be a more generic "The rating is X out of 5 stars" and "There is no rating".

alice-blue:
> Outstanding issues in my commits:
> 
> 1. There's no visual indicator when you hover over a star and click. In `viewgame` there's a "saved" label that appears, but if you click twice you will not know something happens. In `review` and `editgame` (when adding external reviews) it's missing. I think it should turn from yellow to red after clicking.

On the viewgame page, if I click twice, or even multiple times, I still see "Saved," so I'm not sure what you mean when you say "if you click twice you will not know something happens."

It does turn from yellow to red for me, but I don't see the red until I move the cursor away.

In review page, I thought you wanted to wait to save the rating until the review was submitted anyway--is that still the plan?

On editgame, the stars do not show up at all for me until I hover in the area. Maybe because I'm in dark mode?

Review no stars


> 2. All 3 places that use the star control have a "remove rating", and there's some code duplication about removing that link when there's no rating. I wonder if it's better to have is an "X" to the right of the stars, and make it managed by the star control and the click callback.

I think "Remove Rating" is fine. It's unambiguous. X could be interpreted as "close this window/section" or something like that.

> 3. I couldn't get the button focus highlighting of the stars to work, esp. together with the hover highlighting. Maybe `:focus-visible` doesn't play well with the `+` and `~` selector tricks (see TODO.) I might switch to styling them in javascript, which will give better control about the first issue mentioned above.

I was able to navigate to the stars using only the keyboard, I was able to get it to the first star, and then I couldn't get off that star. I'm guessing that's what you are talking about here.

> 4. In `viewgame`, I don't like the label being changed from "Rate it:" to "My Rating:" pushes the "saved" to the next row. Also, should that be "Your Rating:" instead?

I don't think saved being pushed to the next row is a big deal. If you wanted to change it, it seems like something cosmetic that could be improved in a future PR.

We use "My Activity" at the top navbar for space reasons, but here we also say "Have you played this game?" I think it's fine either way ("my rating" or "your rating"), as long as there's enough space.

> 6. In `viewgame`, when submitting the rating after a click, if we failed to contact the server, or it returned an error, I think the user should be alerted. Since it's rare, is it OK to use an `alert()`?

I don't know much about this, but it seems ok to me.

> 7. I kept the "first control" focus in the review page if the accessibility setting is turned on. Can it be deleted? Is it useful?

I don't know anything about this.

alice-blue:
> Can we consult with actual users of this feature, even if it's just to get feedback on the new implementation?

We could ask on intfiction.org to see if anyone wants to give feedback.

salty-horse:
> On the viewgame page, if I click twice, or even multiple times, I still see "Saved," so I'm not sure what you mean when you say "if you click twice you will not know something happens."

There's no visual feedback that the second click did anything, since the first click's visual feedback of "saved" is already on the screen.

> In review page, I thought you wanted to wait to save the rating until the review was submitted anyway--is that still the plan?

This is how it worked previously in the review page. I didn't change anything. Only the "viewgame" page updates the review on the server at the moment you click a star.
 
> On editgame, the stars do not show up at all for me until I hover in the area. Maybe because I'm in dark mode?

I will test. It seems like a bug.

> > 2. All 3 places that use the star control have a "remove rating", and there's some code duplication about removing that link when there's no rating. I wonder if it's better to have is an "X" to the right of the stars, and make it managed by the star control and the click callback.
> 
> I think "Remove Rating" is fine. It's unambiguous. X could be interpreted as "close this window/section" or something like that.

Do you think it will be confusing even if it will be on the same row as the stars, in the same size, and with a label visible when you hover over it?

> > 3. I couldn't get the button focus highlighting of the stars to work, esp. together with the hover highlighting. Maybe `:focus-visible` doesn't play well with the `+` and `~` selector tricks (see TODO.) I might switch to styling them in javascript, which will give better control about the first issue mentioned above.
> 
> I was able to navigate to the stars using only the keyboard, I was able to get it to the first star, and then I couldn't get off that star. I'm guessing that's what you are talking about here.

I only described a problem with visuals, not controls. (setting the focus with the keyboard on the fourth star didn't turn 3 stars to its left yellow.)

What do you mean "couldn't get off the star"? Tab did nothing?

alice-blue:
> What do you mean "couldn't get off the star"? Tab did nothing?

I forget if I tried tab or arrow keys.

salty-horse:
The arrow keys worked (instead of tabs) with the initial radio button implementation, due to how radio button navigation works in most browsers.

My button implementation doesn't work with arrows. I can add that, but Goodreads' implementation also doesn't support it, so maybe it not a must.

alice-blue:
> > On editgame, the stars do not show up at all for me until I hover in the area. Maybe because I'm in dark mode?
> 
> I will test. It seems like a bug.

FYI, there were some recent commits that changed the CSS for dark mode, and I'm not sure if this branch reflects those changes.



> > I think "Remove Rating" is fine. It's unambiguous. X could be interpreted as "close this window/section" or something like that.
> 
> Do you think it will be confusing even if it will be on the same row as the stars, in the same size, and with a label visible when you hover over it?

That might be fine, too.

> The arrow keys worked (instead of tabs) with the initial radio button implementation, due to how radio button navigation works in most browsers.
> 
> My button implementation doesn't work with arrows. I can add that, but Goodreads' implementation also doesn't support it, so maybe it not a must.

Ok--I don't usually use keyboard controls, so I don't know how it usually works.

salty-horse:
I rebased the branch onto `main`, and now the editgame stars appear fine in dark mode.

dfabulich:
OK, I've taken @salty-horse's WIP commit and pushed it as accessible-star-controls-ori-wip, and I attempted to fix the issues you raised using my original radio-button approach I grabbed from https://dev.to/grahamthedev/5-star-rating-system-actually-accessible-no-js-no-wai-aria-3idl.

> 1. Since the radio buttons have no "confirmation" step, just changing the focus over them with the keyboard changes the selection

This no longer happens. Now, when you keydown on a star, a "Submit Rating" button appears, and we refuse to auto-submit a rating while that button is visible.

I note that navigating to stars on iOS Voiceover does _not_ count as "keydown," so you can swipe right to get to the fourth star without submitting changes, and then double tap to activate it, and _that_ auto-submits a change. (I've even added `aria-polite` to the save message, so it announces "Saved" when you select a radio button.)

> 2. If you rated something as 3 stars, then want to change it to 5, It's weird that 3 stars are red, and 2 are yellow

This is now fixed with some tricky CSS selectors.

> 3. After rating something, the "remove it" should show up immediately.

This is now handled entirely with CSS.

> 4. Missing ARIA labels: The fieldset should have an aria-label specifying the current rating

I'm not sure if that's necessary. If these were real radio buttons, we wouldn't ordinarily have a separate label or indicator telling you which one is checked.

> 5. If there was an issue sending the data to the server, I think the UI should revert back to the state before the click

I didn't fix that, but we could try to fix it later.

salty-horse:
Looks good! Tried it in all 3 pages.

That "Remove rating" and "Submit rating buttons is very imposing. Any thoughts on using an X button [like this example](https://www.w3.org/WAI/tutorials/forms/custom-controls/#a-star-rating)? (The example puts it on the left, but if you want to show/hide it as needed, it should be on the right.)

If I tab over the star rating control without changing anything, it still shows a "submit rating" button. I don't think it should.

I'm still missing some visual feedback on mousedown/up. Maybe shade and unshade the current and previous stars under the cursor a bit, then unshade it? Similar to how radio buttons are styled by default.

salty-horse:
I know this is how radio buttons work, but their use on the game page still feels so weird to me. If I change the selection with the keyboard without submitting, it seems that the only way to know what my server-stored current rating is to reload the page. Is this OK?

dfabulich:
> That "Remove rating" and "Submit rating buttons is very imposing. Any thoughts on using an X button [like this example](https://www.w3.org/WAI/tutorials/forms/custom-controls/#a-star-rating)? (The example puts it on the left, but if you want to show/hide it as needed, it should be on the right.)

I think I agree with @alice-blue here:

> @alice-blue: I think "Remove Rating" is fine. It's unambiguous. X could be interpreted as "close this window/section" or something like that.

We raised the idea of gathering feedback. I think the thing to do is just to ship it to ifdb.org and then post for public comment on intfiction.org. I'm pretty sure we don't have any active screen-reader users in the IFDB Advisors group. (We can always change it or roll it back later.)

> If I tab over the star rating control without changing anything, it still shows a "submit rating" button. I don't think it should.

Fixed! Now only arrow keys will activate the button.

> I'm still missing some visual feedback on mousedown/up.

Fixed! Now the yellow stars turn slightly darker when you mousedown.

dfabulich:
Based on conversation here https://intfiction.org/t/changes-on-deck-on-dev-ifdb-org-2024-10-18-perf-star-ratings/71699 I fixed a bug where it would complain if you try to "Remove Rating" when you have a written review (now it just allows it), and I made the "Remove Rating" button autosubmit even if you previously used your keyboard to select a rating.

dfabulich:
Seems like the consensus on the forum is to open this up to public discussion.

salty-horse:
@dfabulich I copied `accessible-star-controls-ori-wip` to my own repository, so it can be removed from this one if you wish. I don't see a reason to keep it around.

PR #352, opened 2024-10-05
Return tag synonyms in game tag search results

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/73

For example, as I write this, [Zeppelin Adventure](https://ifdb.org/viewgame?id=op5a8ug80gwyoagl) is tagged "sci-fi" but not "science fiction". When I search for "[adventure tag:science fiction](https://ifdb.org/search?searchbar=adventure+tag%3Ascience+fiction)" on ifdb.org, Zeppelin Adventure doesn't show up, but in this branch, it _does_ show up, and in search results, it says: "tags matched: sci-fi"

image

There's no UI to edit tag synonyms, but that can come later, assuming I've got the schema right.

Am I right in thinking we want to allow synonyms to be asymmetrical? So searches for tag X return games tagged X or Y, but searches for tag Y don't necessarily return games tagged X? (That's why my `tagsynonyms` table has a `fromtag` and `totag` column.)

### DIFF
--- a/sql/patch-schema-2.sql
+++ b/sql/patch-schema-2.sql
@@ -3,3 +3,13 @@ USE ifdb;
 -- use this script for pending changes to the production DB schema
 
 
+CREATE TABLE `tagsynonyms` (
+    `tagsynonymid` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+    `fromtag` varchar(255) COLLATE latin1_german2_ci NOT NULL,
+    `totag` varchar(255) COLLATE latin1_german2_ci NOT NULL,
+    PRIMARY KEY (`tagsynonymid`),
+    KEY `fromtag` (`fromtag`),
+    UNIQUE KEY `from_to` (`fromtag`, `totag`)
+) ENGINE = MyISAM DEFAULT CHARSET = latin1 COLLATE = latin1_german2_ci;
+
+insert into tagsynonyms (fromtag, totag) values ('science fiction', 'sci-fi');
--- a/www/search
+++ b/www/search
@@ -1458,11 +1458,16 @@ else if ($term || $browse)
                     $tagsMatched = array();
                     foreach ($specials as $s) {
                         if ($s[0][0] == "tags") {
+                            $result = mysqli_execute_query($db, 'select totag from tagsynonyms where fromtag = ?', [$s[1]]);
+                            $synonyms = array_merge(...mysqli_fetch_all($result, MYSQLI_NUM));
+                            $matchables = array_merge([$s[1]], $synonyms);
                             foreach ($tags as $t) {
-                                if (strpos(strtolower($t),
-                                           strtolower($s[1])) !== false) {
-                                    $tagsMatched[$t] = true;
-                                    break;
+                                foreach ($matchables as $m) {
+                                    if (strpos(strtolower($t),
+                                               strtolower($m)) !== false) {
+                                        $tagsMatched[$t] = true;
+                                        break;
+                                    }
                                 }
                             }
                         }
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -1051,6 +1051,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
 
     // Build tags join
     $tagsJoin = "";
+    $bindParameters = [];
     if ($tagsToMatch) {
         // Limit number of tags to avoid abuse
         $tagsToMatch = array_slice($tagsToMatch, 0, 20);
@@ -1064,7 +1065,9 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                 $t .= " on t0.gameid = t$i.gameid";
             }
             $tagsTables[] = $t;
-            $tagsWhereParts[] = "t$i.tag = ?";
+            $tagsWhereParts[] = "t$i.tag in (select ? union select totag from tagsynonyms where fromtag = ?)";
+            $bindParameters[]= $tagsToMatch[$i];
+            $bindParameters[]= $tagsToMatch[$i];
         }
         $tagsSubJoin = implode(" join ", $tagsTables);
         $tagsWhere = implode(" and ", $tagsWhereParts);
@@ -1127,11 +1130,13 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
               $baseOrderBy
             $limit";
 
+    $bindParameters = array_merge($bindParameters, $tagsToNegate);
+
     if ($logging_level) {
         error_log($sql);
+        error_log(json_encode($bindParameters));
     }
 
-    $bindParameters = array_merge($tagsToMatch, $tagsToNegate);
 
     // run the query
     $result = mysqli_execute_query($db, $sql, $bindParameters);


dfabulich:
@alice-blue

www/search
```diff
@@ -1458,11 +1458,16 @@ else if ($term || $browse)
...
                         if ($s[0][0] == "tags") {
+                            $result = mysqli_execute_query($db, 'select totag from tagsynonyms where fromtag = ?', [$s[1]]);
+                            $synonyms = array_merge(...mysqli_fetch_all($result, MYSQLI_NUM));
+                            $matchables = array_merge([$s[1]], $synonyms);
                             foreach ($tags as $t) {
-                                if (strpos(strtolower($t),
-                                           strtolower($s[1])) !== false) {
-                                    $tagsMatched[$t] = true;
-                                    break;
+                                foreach ($matchables as $m) {
```
salty-horse, line 1465: I don't know how much data this is supposed to have, but here's an optimization suggestion, in case it's useful:

Move the `strtolower` out of the nested loop. Compute it once for `$t` and `$m`. This will need a mapping from the lowercased strings to their original form. (Need to be careful of duplicates, like "aa" mapping to both "AA" and "Aa", but maybe this should be enforced when saving new synonyms.)

If you don't mind importing another package, you can even use [set intersection](https://www.php.net/manual/en/class.ds-set.php) instead of the loop.

salty-horse (commented):
Didn't run the code, but it seems OK

alice-blue:
It would be easier to experiment with this once some synonyms that are already marked, but I'm wondering if it's still going to be possible to search and without returning synonyms, in case someone wants to. I'm not sure of a use case for it, but I know I've run into e.g. a website trying to automatically adjust my search in a way that is actively unhelpful.

alice-blue:
> Am I right in thinking we want to allow synonyms to be asymmetrical? So searches for tag X return games tagged X or Y, but searches for tag Y don't necessarily return games tagged X? (That's why my `tagsynonyms` table has a `fromtag` and `totag` column.)

If we're going to have "official" tags and then "unofficial" synonyms that redirect to the official tags (kind of like a redirect page on a wiki, which exists just to send you to the correct page), then yes. If we're thinking of using this for tag suggestions, then I'm guessing we are going to want to suggest official tags if a user types an unofficial synonym, but we aren't necessarily going to want to suggest unofficial synonyms. (Unless there is a separate option for that.)

sql/patch-schema-2.sql
```diff
@@ -3,3 +3,13 @@ USE ifdb;
...
+CREATE TABLE `tagsynonyms` (
+    `tagsynonymid` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+    `fromtag` varchar(255) COLLATE latin1_german2_ci NOT NULL,
+    `totag` varchar(255) COLLATE latin1_german2_ci NOT NULL,
+    PRIMARY KEY (`tagsynonymid`),
+    KEY `fromtag` (`fromtag`),
+    UNIQUE KEY `from_to` (`fromtag`, `totag`)
+) ENGINE = MyISAM DEFAULT CHARSET = latin1 COLLATE = latin1_german2_ci;
+
+insert into tagsynonyms (fromtag, totag) values ('science fiction', 'sci-fi');
```
alice-blue, line 15: Does this mean that if I type `science fiction` into the search bar, it should return `sc-fi`? If so, it did not work for me. I also tried it in the opposite direction, and that didn't work either.

www/search
alice-blue, line 1: A problem message showed up for line 194 on the search page:

line 194

sql/patch-schema-2.sql
dfabulich: Changes to `patch-schema-2.sql` require you to `docker compose down` and then `docker compose up` to rebuild the database before testing. It did work for me. Specifically, Zeppelin Adventure is tagged "sci-fi" and not "science fiction", but when I search for "adventure tag:science fiction", Zepplin Adventure is in the results when I search, and not when I do a search on the main branch.

www/search
dfabulich: Weird. I'm pretty sure this line does work, but I'd merge a PR to fix this.

dfabulich:
> It would be easier to experiment with this once some synonyms are already marked, but I'm wondering if it's still going to be possible to search tags without returning synonyms, in case someone wants to. I'm not sure of a use case for it, but I know I've run into e.g. a website trying to automatically adjust my search in a way that is actively unhelpful.

I'm open to that, but I'm not sure how it should work.

dfabulich:
> If we're going to have "official" tags and then "unofficial" synonyms that redirect to the official tags (kind of like a redirect page on a wiki, which exists just to send you to the correct page), then yes. If we're thinking of using this for tag suggestions, then I'm guessing we are going to want to suggest official tags if a user types an unofficial synonym, but we aren't going to want to suggest unofficial synonyms. (It's possible someone might want to search for all synonyms, though, even unofficial ones, outside the context of tag suggestions.)

I'm not sure I understand how you're thinking about this.

For "Tag Suggestions" https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/36 I was imagining just doing a little search by tags as you type. So if you type "sci-" we'd suggest "sci-fi."

In my mind, that's fine, because the point of tag synonyms in search is so we don't have to actually edit all the games tagged "sci-fi" and tag them "science fiction." We don't even have to nudge people away from "sci-fi" towards "science fiction." We can just leave it alone, and searchers will get the right results anyway.

sql/patch-schema-2.sql
alice-blue: Maybe I'm misunderstanding what you were trying to do. I was on the "tags" tab of the search page:

search tags

When I type in "science fiction," "sci-fi" does not show up in the results. I deleted the previous codespace and started a new one--would that mean I have the latest version of the database?

alice-blue: When I search for `adventure tag:science fiction` on the games tab, I do get Zepplin Adventure in results.

alice-blue:
> In my mind, that's fine, because the point of tag synonyms in search is so we don't have to actually edit all the games tagged "sci-fi" and tag them "science fiction." We don't even have to nudge people away from "sci-fi" towards "science fiction." We can just leave it alone, and searchers will get the right results anyway.

Ok, I was assuming that we were going to want to clean up the tags--i.e. change all "sci-fi" to "science fiction" or vice versa, in an automated way. In the tag table on the showtags page, I think it would be more helpful to see at a glance that there are (for example) 500 science fiction games, rather than 200 "sci-fi" games and then 300 "science fiction" games.

sql/patch-schema-2.sql
dfabulich: I updated the main description of the PR to clarify. I'm trying to make this search <https://ifdb.org/search?searchfor=adventure+tag%3Ascience+fiction> which currently does not return Zeppelin Adventure return Zeppelin Adventure.

dfabulich:
I've reopend this discussion back on the forum. https://intfiction.org/t/tagfest-2024-suggested-tag-consolidations/71371/31?u=dfabulich

dfabulich:
I tentatively think I want to close this PR in favor of #358 and #360.

Given that "sci-fi" is the unwanted tag and that "science fiction" is the official tag, in this PR, I was imagining that you'd search for "science fiction" but still find all the "sci-fi" games as well.

But, in #358, I think we're just going to remove all the "sci-fi" tags, replacing them with "science fiction." So we no longer need any search synonyms.

I say "tentatively" because it _could_ make sense to have someone search for `tag:sci-fi` and find all the "science fiction" games. But I don't think it's worth it.

1. I don't really think anybody will do a search for `tag:sci-fi` after we delete all the sci-fi tags. The vast majority of `tag:` searches come from clicking on a tag on the `viewgame` page; hardly anybody will manually type out `tag:` and play that way.
1. It will add complexity to `searchutil.php`, which is already very complicated performance-sensitive code.
1. I'd have to add a message to the search results page explaining what I did, and why

dfabulich:
Closing this for now; we can reopen it again if we feel the need.

PR #353, opened 2024-10-05
Fix implicit search by TUID

dfabulich:
You're supposed to be able to paste a TUID into IFDB search, with no operator, and have it just work, but we broke it when we removed implicit search by IFID.

### DIFF
--- a/www/searchutil.php
+++ b/www/searchutil.php
@@ -1086,7 +1086,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
             error_log($sql);
             error_log($words[0]);
         }
-        $result = mysqli_execute_query($db, $sql, [$words[0], $words[0]]);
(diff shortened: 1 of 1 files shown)


salty-horse (approved):
Doh. I think this error would have been caught more quickly if that one line of SQL was part of the query call, instead of a variable.

Why not simplify it? Is it that useful to have logging of this very basic query?

salty-horse:
I meant just doing this. The query is too simple to log for debugging purposes, and keeping the actual query string together with the call makes it easy to verify that the number of question marks matches the number of items in the list.
```diff
-        $sql = "select games.id as gameid from games where games.id = ?";
-        if ($logging_level) {
-            error_log($sql);
-            error_log($words[0]);
-        }
-        $result = mysqli_execute_query($db, $sql, [$words[0]]);
+        $result = mysqli_execute_query($db,
+            "select games.id as gameid from games where games.id = ?",
+            [$words[0]]);
```

PR #354, opened 2024-10-06
Allow sorting the tags table in alphabetical order

alice-blue:
People seemed to like the idea of allowing the table of tags to be sorted in alphabetical order: https://intfiction.org/t/thoughts-on-the-table-of-all-tags-on-ifdb-showtags-page/71482

This adds an option for that.

There are links to show the top 20, 50, and 100 tags in table form--whether in alphabetical order or not--but at the moment, those lines are commented out (for the table only, not for the cloud) because I was unsure whether to include them. (It's still possible to list only the top N tags, if you type the limit into the url, but with these lines commented out, that possibility is not advertised).

### DIFF
--- a/www/showtags
+++ b/www/showtags
@@ -161,16 +161,37 @@ if (count($tags) == 0) {
     
         echo "</center></div>";
     } else {
-        echo "<h1>$title</h1>"
-        . "<p>Here are $limitName on IFDB, along with numbers showing how many games have each tag. "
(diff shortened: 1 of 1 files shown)


salty-horse:
Works fine for me!

alice-blue:
Thanks! Should I delete those commented out lines?

dfabulich:
I deleted the commented out lines myself.

PR #357, opened 2024-10-06
Dark mode colors

dfabulich:
# Tag Editor Before

image

# Tag Editor After

image

# Edit Link Before

image

# Edit Link After

image

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -2264,6 +2264,22 @@ span.sfl-save-autodesc {
     font-style: italic;
     padding-left: 1em;
 }
+#tagEditor,
+#tagDeletor {
+    position: absolute;
+    z-index: 10000;
+    top: -5em;
+    left: 0px;
+    display: none;
+    background: #ffffff;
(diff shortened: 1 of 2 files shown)


salty-horse:
Is the "Edit Link" box background supposed to be white? I'd expect it to be black like the page behind it, or a slightly lighter shade.

What about `textarea`s? In github, those have a dark background with gray text.

Github uses `#0d1117` instead of black for the background of the site, and text input boxes - I think it feels less oppressive.

alice-blue:
Related to text areas in dark mode: https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/515

dfabulich:
Yeah, there's definitely more available to fix here. I mostly just wanted to fix the illegible Help/Close/Cancel buttons and the heading color in the title bar of the popup. We can fix other stuff in future PRs.

alice-blue:
Screenshots look ok to me (I did not open in codespace).

PR #358, opened 2024-10-07
Allow moderators to add/delete tags in bulk

dfabulich:
Bulk adding tags probably shouldn't be restricted to moderators; anyone can already add as many tags as they want, and can even write code to do it using the IFDB API. (But I don't really know where I'd put a feature like that. I don't think I'd want to link to it right off the `viewgame` page. Maybe a secret URL?)

Fixes https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/514

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -75,6 +75,57 @@ if (isset($_REQUEST['adminemail'])) {
     exit();
 }
 
+if (isset($_REQUEST['bulkdeletetag'])) {
+    $tag = $_POST['tag'] ?? null;
+    $confirm = $_POST['confirm'] ?? null;
+    pageHeader("Bulk delete tag");
+    if (!$tag) {
+        ?>
+        <h1>Bulk delete tag</h1>
+        <p><form method=post>Delete Tag: <input type=hidden name=bulkdeletetag value=''><input name=tag><button>Search</button></form></p>
+        <?php
+    } else if (!$confirm) {
+        $result = mysqli_execute_query($db, 'select distinct(id), title from games join gametags on games.id=gameid where tag = ? order by title', [$tag]);
+        $rows = mysqli_fetch_all($result, MYSQLI_NUM);
+        if (!count($rows)) {
+            echo "<h1>Bulk delete tag: ". htmlspecialcharx($tag)."</h1>\n"
+                ."<p><span class=errmsg>There are no games with the tag \"".htmlspecialcharx($tag)."\".</span>"
+                ."<p><form method=post>Delete Tag: <input type=hidden name=bulkdeletetag value=''><input name=tag><button>Search</button></form></p>";
+            pageFooter();
+            exit();
+        }
+        echo "<h1>Bulk delete tag: ". htmlspecialcharx($tag)."</h1>\n"
+            ."<p>" . count($rows) . " game(s):</p>";
+        global $nonce;
+        echo "<style nonce='$nonce'>#games { display: grid; grid-template-columns: repeat(auto-fill, 20ch);</style><div id=games>\n";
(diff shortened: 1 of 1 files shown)


alice-blue:
I wasn't sure at first how this is supposed to work. I'm supposed to search for a tag and then add it? I think it could be worded a little more clearly to explain how to use this.

alice-blue:
It might also be nice to have a convenient link back to the list of admin options, because after bulk adding new tags, the admin might want to go back and then bulk delete the old tags.

salty-horse:
I agree that the "bulk delete tags" form is confusing.
I would reverse the fields and label them like so:
> Add this tag _______
> to all games that have this tag ______

But without context, I don't think it's at all clear why someone would want to use this. Maybe both forms should be on the same page, and there should be some text describing the various use-cases where these forms are useful.
If the only use-case is "rename tag", why not have that as a tool, instead?

I think this is too destructive to be available for all users, but if you insist, it can be in an `expertops` page linked to from the Contribute section.

alice-blue:
> the various use-cases where these forms are useful.
> If the only use-case is "rename tag", why not have that as a tool, instead?

I could imagine a few use cases for bulk adding/deleting that would not involve renaming:

- To add an additional tag that's a more general category name. For example, if there are a bunch of games tagged "i7 source" but we also want them all to have the more general tag "source code available."

- To delete objectionable tags that have been added to several games. This could also be done one at a time, but it's probably easier to do it all at once.

That said, I'm not a mod or an admin, so I'm just guessing how this might be used.

But maybe a "rename" shortcut would be nice to have.

dfabulich:
Pushed updates to make it clearer how to use it. Now the "Bulk add tag" feature uses @salty-horse's suggested language.

image

As @alice-blue said, I do think it's independently useful to bulk add and bulk delete without bulk rename. I guess I could add a bulk rename for convenience, but I expect that this will be used very rarely, by people who know what they're doing.

As for undoing the change, I think the long-term thing to do is to implement tag history https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/502

That will give us the confidence to allow anyone to delete tags, safe in the knowledge that anyone can undo the changes. I could probably then also add a feature to bulk undo game edits, perhaps restricted to a specific user / timeframe.

But that seems like a lot of work (more work than I want to do right now).

alice-blue:
> As for undoing the change, I think the long-term thing to do is to implement tag history [iftechfoundation/ifdb-suggestion-tracker#502](https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/502)

I'm not super familiar with API, but I assume you can plug in some kind of code or information to send to IFDB to make it do an automated thing. This wouldn't necessarily be a long-term solution, but what about something like this--when you do a bulk delete, it automatically generates an "UNDO this action via API file" which would have the right code to plug into the API to add the same tags back in that were just deleted? (I'm not saying we have to do something like this right now--just wondering if it would make any sense as a short-term solution).

salty-horse:
Looking at your screenshot, I realize the button shouldn't say "Search" since this is a destructive operation.

dfabulich:
> I think this is too destructive to be available for all users, but if you insist, it can be in an `expertops` page linked to from the Contribute section.

FYI, someone did create a script to do Bulk Add. https://intfiction.org/t/announcing-the-ifdb-tag-manipulator/71710

[I responded](https://intfiction.org/t/announcing-the-ifdb-tag-manipulator/71710/4?u=dfabulich):

> This script works, and maybe it’s a good thing, but it creeps me out, because if someone does use it in a way they regret, it could be hard for us to clean up.

PR #359, opened 2024-10-07
Tag autocomplete

dfabulich:
image

### DIFF
--- a/www/showtags
+++ b/www/showtags
@@ -24,8 +24,6 @@ else {
     $title = "Top $limit Tags";
 }
 
-pageHeader($title);
-
 // get the top (most frequently used) $limit tags
 $result = mysql_query(
     "select
@@ -99,10 +97,24 @@ function calcDistBin($val, $bins)
 if (count($tags) == 0) {
     echo "<i>There are currently no tags in the database</i>";
 } else {
+    $datalist = get_req_data("datalist");
 
     $cloud = get_req_data("cloud");
 
-    if ($cloud) {
+    if ($datalist) {
+        header('Content-Type: text/xml');
+        header('Cache-Control: public, max-age=3600');
+        echo "<datalist id=\"tags-list\">\n";
+        foreach ($tags as $t) {
+            $freq = $t['freq'];
+            if ($freq < 10) continue;
+            $tag = $t['tag'];
+            echo "  <option value=\"".htmlspecialcharx($tag) . "\"></option>\n";
+        }
+        echo "</datalist>\n";
+        exit();
+    } else if ($cloud) {
+        pageHeader($title);
         $i = $cnt - 1;
         $bins = [(int)($i * 0.02),     // bottom 2% in bin 1
                     (int)($i * 0.07),  // next 5% in bin 2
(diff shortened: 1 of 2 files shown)


alice-blue:
Looks nice! I like how the completion is not limited only to tags that start with what you've typed  (e.g. "I7 source available" might show up if you type "source," which is good.)

Is the tag supposed to show the frequency in parentheses in the Edit Tag box even after you've already clicked "add"?

tag editor with frequency

alice-blue:
I got an error when I tried to save changes:

401 error

dfabulich:
> Is the tag supposed to show the frequency in parentheses in the Edit Tag box even after you've already clicked "add"?

Oops! I had to fix that by removing the frequency. Should be fixed in the latest version of the PR.

401 error means that you're not logged in. Try refreshing. There's an error message returned by the backend, but we don't show it. https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/483 (You can also look at the Network tab in Chrome Dev Tools to read the error message.)

alice-blue:
> I had to fix that by removing the frequency. Should be fixed in the latest version of the PR.

It works now. Too bad about having to remove the frequency, though. When someone is choosing between similar tags, it would be nice to be able to see how common they are. I wonder if it would be possible to show a count if you hover over on one of the options or focus on it with the keyboard? Or, if more common tags are listed first in the autocomplete list, maybe that could be mentioned somewhere? Or maybe as we clean up tags, it'll become less of an issue.

alice-blue:
I notice that if I click inside the text input field, it lists suggestions even if I haven't typed anything yet. Is that supposed to happen? Is that list showing all of the tags that appear at least 10 times?

www/showtags
```diff
@@ -99,10 +97,24 @@ function calcDistBin($val, $bins)
...
     $cloud = get_req_data("cloud");
 
-    if ($cloud) {
+    if ($datalist) {
+        header('Content-Type: text/xml');
+        header('Cache-Control: public, max-age=3600');
+        echo "<datalist id=\"tags-list\">\n";
+        foreach ($tags as $t) {
+            $freq = $t['freq'];
+            if ($freq < 10) continue;
```
salty-horse, line 110: this check can be part of the SQL above. In the test DB it will fetch 861 records instead of 4937.

It can be simple `having freq >= ?` and then the query param is just `$datalist ? 10 : 0`

dfabulich: done!

alice-blue:
> IMO `<datalist>` is "good enough" for a first draft.

It might be good enough long term! I don't think this is a big deal. I think people will be happy to be able to start using it.

www/showtags
```diff
@@ -24,17 +24,16 @@ else {
...
+$result = mysqli_execute_query($db,
     "select
        tag, count(tag) as freq,
        (unix_timestamp(now()) - unix_timestamp(max(moddate))) as age
      from gametags
      group by tag
+     having freq > ?
      order by freq desc, tag
-     $limitClause", $db);
+     $limitClause", [$datalist ? 10 : 0]);
```
salty-horse, line 36: `$datalist` is undefined here. Move its initialization up.

PR #360, opened 2024-10-08
Ban tag synonyms

dfabulich:
In this PR, we populate a table of `blockedtagsynonyms`. We forbid anyone to tag a game with a tag on this list, and we recommend that the user use a synonymous tag instead, e.g. we'd block "sci-fi" and tell people to use "science fiction" instead.

Along the way, I had to fix https://github.com/iftechfoundation/ifdb/issues/396 so I could show an error message at all.

Notably, when tagging fails, the changes remain visible in the page, as if they'd kinda succeeded, with the red error message "Not Saved."

image

This is because the user might have submitted multiple tags X, Y, and Z, but only tag Y is blocked. If we just revert all the changes, we'd have to discard X and Z. The user can fix the issue by clicking "Edit," unchecking Y, and submitting that, or by refreshing the page, discarding all of the tags they submitted.

### DIFF
--- a/sql/patch-schema-2.sql
+++ b/sql/patch-schema-2.sql
@@ -3,3 +3,13 @@ USE ifdb;
 -- use this script for pending changes to the production DB schema
 
 
+CREATE TABLE `blockedtagsynonyms` (
+    `blockedtagsynonymid` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+    `blockedtag` varchar(255) COLLATE latin1_german2_ci NOT NULL,
+    `preferredtag` varchar(255) COLLATE latin1_german2_ci NOT NULL,
+    PRIMARY KEY (`blockedtagsynonymid`),
+    UNIQUE KEY `blockedtag` (`blockedtag`)
+) ENGINE = MyISAM DEFAULT CHARSET = latin1 COLLATE = latin1_german2_ci;
+
+insert into blockedtagsynonyms (blockedtag, preferredtag)
+values ('sci-fi', 'science fiction');
--- a/www/adminops
+++ b/www/adminops
@@ -185,6 +185,67 @@ if (isset($_REQUEST['bulkdeletetag'])) {
     exit();
 }
 
+if (isset($_REQUEST['blocktag'])) {
+    $tag = $_POST['blockedtag'] ?? null;
+    $preferredtag = $_POST['preferredtag'] ?? null;
+    $deleteblock = $_POST['deleteblock'] ?? null;
+    pageHeader("Block tag");
+    if ($deleteblock) {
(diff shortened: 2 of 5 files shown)


dfabulich:
This PR #360 seems superficially similar to #352, but they're totally different. In #352, users who search for tag X will get search results containing tag Y. In this PR #360, users are unable to add certain tags that have preferred synonyms.

alice-blue:
I'm not sure if I posted my comment in the right spot, so I'll add a link to it here: https://github.com/iftechfoundation/ifdb/commit/f82e33d0557239a9d3257d0999d87e886b5df4fa#commitcomment-147682518

dfabulich:
Yeah, that comment is directly on the commit itself, rather than on the commit in the context of the PR.

In general:
* When commenting on the PR as a whole ("could we do it on add instead of on save?") that can be a comment right here, directly on the PR's thread of comments
* When commenting on lines of code in the PR, click on the Commits tab from the PR page, then click on a commit there. The URL will be like <https://github.com/iftechfoundation/ifdb/pull/360/commits/2406e8e8633f6f3e6cac1a221df77cbe9bc4dece> (including `/pull/###`) instead of <https://github.com/iftechfoundation/ifdb/commit/f82e33d0557239a9d3257d0999d87e886b5df4fa> (with no PR ###)

As for the substance of your comments:

> When I tried saving tags with the "Test Tester" account, it told me "Your new user account is still pending review. Editing is not available until the account has been approved."

Yeah, the way to handle that is to login as the admin, then navigate to the System Maintenance Panel and navigate to Test Tester in the User list. There, you can change Test Tester's profile status to "Normal."

> Would it be possible to intercept the tag right when the user clicks "add" instead of waiting until the user clicks "save"?

I agree that it would be better to check for blocked tags when adding rather than waiting until saving, and it would be possible, but it would be more work. The fix for this was just:

1. Fix https://github.com/iftechfoundation/ifdb-suggestion-tracker/issues/483
2. Define the `blockedtagsynonyms` table and put a row in it
3. Ten lines of code to return an error message if the user tries to use any blocked tags.

> Also, I think I would word the error message a little differently....as-is, it might come across as scolding the user for doing something wrong. I would prefer something like `To help keep IFDB's tags tidy, please use the existing tag "science fiction" instead of "sci-fi."`

Done!

> Question: how will new preferred synonyms get added to the table?

As in #352, I didn't yet make a UI to edit tag synonyms, but that can come later. In the meantime, once this PR merges, I can run SQL directly on the database to define blocked synonyms.

alice-blue:
Thanks!

(Another possible thing to test is what happens if someone enters more than one blocked synonym. I'm wondering if there would be multiple error message pop-ups.)

dfabulich:
There will only be one, for the first one the user submits.

dfabulich:
Added admin UI to block tag synonyms.

PR #895, opened 2024-10-18
Enable bfcache on session pages

dfabulich:
`session_start()` automatically tosses in a `session.cache_limiter` setting, "nocache" that sets "Cache-Control: no-store, no-cache, must-revalidate"

https://www.php.net/manual/en/session.configuration.php#ini.session.cache-limiter
https://www.php.net/manual/en/function.session-cache-limiter.php

That's normally excessive, but harmless. `private, no-cache` would be sufficient to force revalidation.

But, it turns out, `no-store` breaks bfcache. https://web.dev/articles/bfcache#minimize-no-store

So, here, we're disabling the `cache_limiter` and setting an explicit `Cache-Control: private, no-cache`. This enables bfcache, which is especially snappy when you do a search, click on a game, and click the back button.

### DIFF
--- a/www/session-start.php
+++ b/www/session-start.php
@@ -21,6 +21,11 @@ session_set_cookie_params([
     'samesite' => 'Lax'
 ]);
 
+// session_start automatically sets a cache-control header by default; disable that
+session_cache_limiter('');
+
+header('Cache-Control: private, no-cache');
+
 @session_start();
 
 ?>


salty-horse:
Sorry, I'm not familiar with caching, and don't know the implications.

dfabulich:
Well, let me write a few more words here, and then we can decide what a good next step is.

First, you can see the impact of this by viewing localhost:8080 in browser Dev Tools in the Network tab.

* When loading the home page on the main branch, the response Cache-Control header says, "no-store, no-cache, must-revalidate"
* When loading the home page on this PR branch, it says "private, no-cache"

So, then, what's the difference?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control tries to explain, but it's kinda encyclopedic.

One of the first things to note about Cache Control is that there are multiple layers of "caching" that it's intended to control. 

1. **Skip request:** if a cache entry is "fresh", the browser can simply skip requesting it, and use the version from the cache.
2. **Revalidation with Conditional GET:** The browser can make a conditional request for the file, including the `If-Modified-Since` request header (matching the `Last-Modified` response header) and the `If-None-Match` request header (matching the `Etag` response header). A conditional get can include either `If-Modified-Since` or `If-None-Match` or both.

    The server can respond to a conditional get with HTTP status code "304 Not Modified." For short responses, 304 responses save almost no time at all, because almost all of the cost  was the time spent initiating the request. But for large responses (big images, media files, etc.) 304 can save you megabytes of transfer.
    
    Note that browsers will revalidate the document itself when you click the refresh buttton, but normally won't revalidate "subresources" (CSS/JS) when you refresh. You can do a shift-refresh to do a "hard refresh," which revalidates subresources, too, but only web developers know how to do that (and you can't do it at all on a phone).
3. **Shared Proxies:** You can use a shared HTTP proxy, either a "forward proxy" where the client funnels all of their requests through it (typical at some companies/schools), or a "reverse proxy," controlled by the webserver, where the proxy sits in front of a webserver and handles all incoming requsets. Cloudflare is a popular reverse proxy.

    Shared proxies are creepy for two reasons. First, it's possible to accidentally cache data that belongs to user X and then show that data to user Y, if you mess up your cache headers. Second, there's normally no way for an end user to click a button to clear a proxy's cache, not even if the user uses browser settings to clear their own personal cache.

**Cache-Control header directives are very crufty.** Each time a new Cache-Control keyword has been added, there's a backwards compatibility problem with existing clients and proxies who don't understand the directive.

With that said, here's the gist of the directives we think about.

* `max-age`: Sets an age in seconds, after which the response is "stale." You might think that `max-age=0` would block all caching, but that's not so. When a cache entry is stale, you can still revalidate it with a conditional request. `max-age=0` only prevents skipping the request. Furthermore, some historical clients were in the habit of using stale cache entries when the client is offline, skipping the request even though it's stale. That's pretty rare in the 21st century, but a lot of guides on caching were written very defensively.
* `must-revalidate`: This instructs those rare browsers that try to use stale cache entries when offline that, no, when it's stale, you must revalidate it.
* `no-cache`: This one is terribly named, because it just means `max-age=0, must-revalidate`. So, with `no-cache` you might still keep stuff in the cache and use it quite a lot, getting 304 responses.
* `no-store`: This one really means "don't cache at all." A no-store result won't be used for conditional GET, won't be kept in any proxy, etc.
* `private`: This is a directive for proxies: it says that the value can't be shared between users. In practice, shared proxies simply don't store `private` data, and just let the client browsers cache stuff.

So, let's look at what the main branch did: `no-store, no-cache, must-revalidate`. That's pretty redundant; `no-store` should be the last word here. But some 20th century clients don't know what `no-store` means, so it's popular to slap on extra older cache directives just in case. If you don't know what `no-store` means, hopefully you'll know what `no-cache` means, and, barring that, you'll know what `must-revalidate` means.

In this branch, I'm proposing to switch it to `private, no-cache`. That tells shared proxies not to cache it, but it does tell clients that they could, in principle, try to revalidate it. (I could add `must-revalidate` here, but `no-cache` means `must-revalidate`, and I'm not aware of any buggy browser we'd need to support here.)

Now, in point of fact, the home page doesn't have a `Last-Modified` header or an `Etag` header, so `private, no-cache` does effectively the same thing as `no-store`. Proxies won't store it, and browsers will never use it.

But there is _one_ thing that's good for, and that's the bfcache, which is the point of this PR. Browsers have made the call that `no-cache` responses _can_ go in the bfcache, but `no-store` responses can't.

So, in this PR, I'm replacing `no-store, no-cache, must-revalidate` (which just means `no-store`) with an equivalent header `private, no-cache` (with no `Last-Modified` or `Etag` header), just to turn the bfcache on.

# Now what?

We could say:

1. Now that you know the deal, you "approve" the PR, we merge it, and ship it.
2. Or we could say "well, Dan _thinks_ he knows, and he's written a bunch of words about this, but what if he's wrong? We need someone who knows more about caching to double-check that he's right."
3. Or… something else. I could probably put it up on dev and let people click around, but if there's a caching bug here, it's going to be very subtle.

WDYT?

salty-horse:
2 - I suggest asking an expert.

It seems like it's not making the current situation worse, in order to enable a good browser feature, but in general, wouldn't we want to cache some simple pages, where it's easy to calculate Last-Modified, like `/showimage`?

dfabulich:
I don't suppose @curiousdannii would be able to weigh in here?

curiousdannii:
Sorry, I really don't know this much about caching.

But longer caching for images certainly makes sense.

dfabulich:
Maybe @erkyrath could weigh in?

dfabulich:
Eh, I trust myself on this one. I'll roll it back if we need to.

PR #898, opened 2024-10-18
Speculation rules

dfabulich:
https://developer.chrome.com/docs/web-platform/prerender-pages

This makes clicking through on search results instantaneous, as well as clicking on reviews and new games on the home page.

For non-eager results, they're prerendered when you hover or mousedown, which can save a fraction of a second, which is nice, but not _as_ nice as loading them in advance.

### DIFF
--- a/www/newitems.php
+++ b/www/newitems.php
@@ -425,7 +425,7 @@ function showNewItemList($db, $items, $first, $last, $showFlagged, $allowHiddenB
                     echo "<i>\"$summary\"</i>";
 
                 if ($trunc)
-                    echo " - <a href=\"viewgame?id={$r['gameid']}"
+                    echo " - <a class=eager href=\"viewgame?id={$r['gameid']}"
                         . "&review={$r['reviewid']}\">See full review</a>";
 
                 echo "</span></div>";
@@ -500,7 +500,7 @@ function showNewItemList($db, $items, $first, $last, $showFlagged, $allowHiddenB
             // summarize this game
             echo "<div class=\"new-game\">"
                 . "A new listing for "
-                . "<a href=\"viewgame?id={$g['id']}\"><b><i>"
+                . "<a class=eager href=\"viewgame?id={$g['id']}\"><b><i>"
                 . output_encode(htmlspecialcharx($g['title']))
                 . "</i></b></a>, by "
                 . output_encode(htmlspecialcharx($g['author']))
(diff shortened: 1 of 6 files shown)


salty-horse:
Does this have a chance to overburden the server? I guess not, since it's max +5 page loads per search.

www/search
```diff
@@ -1508,13 +1512,15 @@ else if ($term || $browse)
...
+                $eagerness = ($i < 5) ? "eager" : "";
+
                 // show the item
                 echo "<p>";
                 if ($art) {
                     // display with art
                     echo "<table class=grid border=0 cellspacing=0 cellpadding=0>"
                         . "<tr><td>"
-                        . "<a href=\"viewgame?id=$id\">"
+                        . "<a class='$eagerness' href=\"viewgame?id=$id\">"
```
salty-horse, line 1523: mix of single and double-quotes in HTML (harmless, similar issue in other lines)

www/newitems.php
```diff
@@ -425,7 +425,7 @@ function showNewItemList($db, $items, $first, $last, $showFlagged, $allowHiddenB
                     echo "<i>\"$summary\"</i>";
 
                 if ($trunc)
-                    echo " - <a href=\"viewgame?id={$r['gameid']}"
+                    echo " - <a class=eager href=\"viewgame?id={$r['gameid']}"
```
salty-horse, line 428: mix of unquoted and quoted HTML attributes (harmless, similar issue in other lines)

dfabulich:
> I guess the only thing that's preloaded is the HTML of the page - is there a way to make it also preload the cover art?

It does way more than that! Speculation Rules offers two modes, `prefetch` and `prerender`. We're using the `prerender` mode, which means "open the page in an invisible tab, downloading all subresources just as if the tab was visible." ("prefetch" just downloads the HTML.)

Then, when you click on the link, it instantly makes the tab visible. The performance is shockingly fast.

salty-horse:
The image is slow to load for me. Since the pages are light on other resources, why not prerender them?

dfabulich:
I don't reproduce that. Here's what I do:

1. Use Google Chrome. Speculation Rules are only supported on Chrome or Chromium-based browsers. https://developer.mozilla.org/en-US/docs/Web/API/Speculation_Rules_API#browser_compatibility 
2. Navigate to a search or browse page. https://dev.ifdb.org/search?browse
3. Click on the first search result. It should display instantly™, including its cover art.

You can verify that the prerendering is working by opening Chrome Dev Tools on the Application tab. In the side bar, near the bottom, there's a section, "speculative loads."

image

It should say that there were "5 success," and you can view the speculations to see which ones are ready.

You can also use the Network tab to set it to 3G throttling. When I do that, and refresh the Counterfeit Monkey page, the page takes 10+ seconds to load, but when I click the Browse page and wait a five seconds before clicking on Counterfeit Monkey, the page pops up instantly, including its cover art.

salty-horse:
OK, I think I might have visited the game pages too quickly. If I wait for the success confirmation in the inspector it works fine.

PR #920, opened 2024-10-24
Add "View Contents" link to IF Archive zips

### LINKED ISSUE #399: Unbox file listing link for every eligible zip (etc) file
IFDB has enough information to provide a link to the Unbox file listing for every (IF-Archive-hosted) zip/tgz/etc file listed in the "external links" section, but at the moment it's only used for the "Play on-line" link on the topmost link that's marked as playable. Unbox's ability to browse zip files is not exposed.

Motivation:
I find myself writing "Contains story file, manual, walkthrough, and source code" in the descriptions of zip files on IFDB, but there's no easy way for IFDB users to reach all those goodies; they can play the story file online, or download the zip and use their own tools to dissect it.
Also, in some cases, there isn't a single "main file" to point to in the zip (the case at the top of my mind is old 8-bit games which come in two parts). Or there are multiple zips with more or less equal claim to people's attention (e.g., for different platforms).

This could add more clutter to an already space-constrained region (cf iftechfoundation/ifdb#403), so it might need to be a slightly gnomic icon with an explanatory 'look inside' tooltip, next to each zip file's link.

dfabulich:
Fixes #399

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1178,10 +1178,17 @@ if (count($links) == 0 && $dlnotes == "") {
         echo "</td><td><div class='downloaditem' id='dlItem$i'>
               <a href=\"$url\"><b>$linktitle</b></a>$zipstar<br>";
 
-        if ($compPri)
+        $unbox_supported = [17, 22, 23, 41];
+
+        if (in_array($fmtcomp, $unbox_supported) && (strpos($url, 'https://ifarchive.org') === 0)) {
+            echo "<div class='details'><a href='https://unbox.ifarchive.org/?url=" . urlencode($url) . "'>View Contents</a></div>";
+        }
+        
+        if ($compPri) {
             echo "<span class=dlzipmain><img src=\"/img/blank.gif\" "
                 . "class=\"zip-contents-arrow\">Contains "
                 . "<b>$compPri</b></span><br>";
+        }
 
         if ($linkdesc) {
             $linkdesc = preg_replace(


www/viewgame
```diff
@@ -1178,10 +1178,17 @@ if (count($links) == 0 && $dlnotes == "") {
         echo "</td><td><div class='downloaditem' id='dlItem$i'>
               <a href=\"$url\"><b>$linktitle</b></a>$zipstar<br>";
 
-        if ($compPri)
+        $unbox_supported = [17, 22, 23, 41];
+
+        if (in_array($fmtcomp, $unbox_supported) && (strpos($url, 'https://ifarchive.org') === 0)) {
```
salty-horse, line 1183: IF Archive links in the database aren't normalized. I see `http://` and `www`, so this prefix test doesn't currently catch everything.

I suggest normalizing them when saving, and updating the database separately.

dfabulich: We do normalize them! At line 1155, we run `urlToMirror`, which used to allow users to select an IF Archive URL, but now it just runs this regex:

```
$result = preg_replace('!^https?://((www\.|mirror\.)?ifarchive(\.[a-z]+?)?\.org)/!', 'https://ifarchive.org/', $url);
```

salty-horse: Nice! BTW, that regex is repeated in what appears to be a similar unboxing operation in line 933:
```php
$unboxUrl = false;
if ($link['compression'] && $link['compressedprimary']) {
    $open = urlencode($link['compressedprimary']);
    if (preg_match('!^https?://(www\.|mirror\.)?ifarchive\.org/!', $link['url'])) {
        $linkUrl = urlencode($link['url']);
        $unboxUrl = "https://unbox.ifarchive.org/?url=$linkUrl&open=$open";
```

salty-horse:
There are several places where the URL isn't normalized. I'll try to fix it.

PR #921, opened 2024-10-25
Show creation date on poll pages

### LINKED ISSUE #423: Show creation date on poll pages
It would be great if the main page for a poll shows the date that poll was created. Right now the only way I know of to see the date is on the "browse polls" page (https://ifdb.org/search?browse&poll&sortby=new); it would be nice if it was shown on each poll's page as well, because it's helpful to know if a poll is recent or years old.

dfabulich:
Fixes #423

### DIFF
--- a/www/poll
+++ b/www/poll
@@ -1450,6 +1450,8 @@ function expandVote(id)
                 . "<i>You created this poll on $created</i> - "
                 . "<a href=\"poll?id=$id&editPoll\">Modify</a> | "
                 . "<a href=\"poll?id=$id&delPoll\">Delete</a></p>";
+        } else {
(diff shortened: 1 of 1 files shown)


www/poll
```diff
@@ -1450,6 +1450,8 @@ function expandVote(id)
                 . "<i>You created this poll on $created</i> - "
                 . "<a href=\"poll?id=$id&editPoll\">Modify</a> | "
                 . "<a href=\"poll?id=$id&delPoll\">Delete</a></p>";
+        } else {
+            echo "<p><hr class=dots><p><i>Poll created on $created</i></p>";
```
salty-horse, line 1454: two `<p>` tags and one `</p>` tag

Issue #926, opened 2024-10-25
Use a CDN


dfabulich:
IFDB doesn't have a CDN, e.g. Cloudflare.

That would allow us to cache cover art, JS, and CSS at the edge, which can improve page-load performance, and protect us from DDOS attacks.

salty-horse:
If JS and CSS are cached, it would be useful to move some code into it :) #914

salty-horse:
Do CDNs support pre-resizing images? IFDB uses several different sizes.

dfabulich:
The easy way is just to set cache headers on the resized images. Then the CDN will cache whatever images we generate.

But, yes, I think Cloudflare has a thing for this.

dfabulich:
Now we use Cloudflare

PR #64, opened 2021-03-10
User authored games RSS feed fix

Durafen:
Fix for rss [user games feed](https://github.com/iftechfoundation/ifdb/issues/62).

PR #927, opened 2024-10-25
Giant search pages

dfabulich:
I'm not sure how I feel about this one.

@alice-blue suggested in a forum thread that it might be easier to browse if the page size was larger.

Search/browse results currently show 20 items, which _does_ seem kinda small. There's an existing feature where if there are fewer than 500 search results, a "Show All" button appears. <s>So, I figured, why not try showing 500 items per page?</s>

EDIT: At @salty-horse's suggestion, I cranked it down to 100.

This changes the size of the `/search?browse` page from 5.7KB (34KB uncompressed) to 14KB (135KB uncompressed). But each cover art image is ~100KB, so maybe that's not really a big deal.

When I enable 20x CPU throttling + 3G network throttling in Chrome Dev Tools, on the main branch I see a 0.9s [Largest Contentful Paint](https://web.dev/articles/optimize-lcp) (LCP), and a 1.4s DOMContentLoaded (DCL) on my local dev environment.

On this branch, I see a 1.3s LCP testing, and a 5.45s DCL. I think LCP is barely affected because the HTML is streaming in, so the browser can start rendering search results and downloading images as soon as the first chunk of HTML is downloaded.

I also added lazy-loading of cover art, to ensure that users wouldn't accidentally download 500 pieces of cover art. Now, with lazy loading, we only start downloading the art as you scroll down through the page.

I have a hunch that we maybe shouldn't turn this on right now, but it might be OK to do it if we use a CDN #926.

Other related ideas:
* infinite scrolling, so subsequent pages of search results pop in, rather than having to click Next Next Next to get to subsequent pages
* Make the first page 20 items for people just exploring the site, but make the 2nd+ page 500 items (for people who want to dig deep)

### DIFF
--- a/www/search
+++ b/www/search
@@ -248,7 +248,7 @@ else {
 
 $pg = isset($_REQUEST['pg']) ? $_REQUEST['pg'] : "";
 $pgAll = false;
-$perPage = 20;
+$perPage = 100;
 $showAllMaxRows = 500;
 $sortby = isset($_REQUEST['sortby']) ? $_REQUEST['sortby'] : "";
 
--- a/www/util.php
+++ b/www/util.php
@@ -2962,7 +2962,7 @@ function coverArtThumbnail($id, $size, $version, $params = "") {
     }
     global $nonce;
     return "<style nonce='$nonce'>.coverart__img { max-width: 35vw; height: auto; }</style>"
-        ."<img class='coverart__img' srcset=\"$thumbnail{$size}x$size$params, $thumbnail{$x15}x$x15$params 1.5x, $thumbnail{$x2}x$x2$params 2x, $thumbnail{$x3}x$x3$params 3x\" src=\"$thumbnail{$size}x$size$params\" height=$size width=$size border=0 alt=\"\">";
+        ."<img class='coverart__img' loading='lazy' srcset=\"$thumbnail{$size}x$size$params, $thumbnail{$x15}x$x15$params 1.5x, $thumbnail{$x2}x$x2$params 2x, $thumbnail{$x3}x$x3$params 3x\" src=\"$thumbnail{$size}x$size$params\" height=$size width=$size border=0 alt=\"\">";
 }
 
 // ----------------------------------------------------------------------------


alice-blue:
I don't know all the implications of changing the page size to 500, but other options might be
* increasing the default number of results per page, but not to 500. (50? 100?)
* adding a link to allow the user to show more results per page (either "show all" or just a larger number than 20), so that it's opt-in rather than default.

www/search
```diff
@@ -1415,7 +1415,7 @@ else if ($term || $browse)
         echo "</div>";
 
         $showFlagged = isset($_GET['showFlagged']) && $_GET['showFlagged'];
-        if (!$showFlagged && $searchType == "game") {
+        if (!$showFlagged && $searchType == "game" && !$browse) {
```
salty-horse, line 1418: why disable this when "browse"ing?

salty-horse:
100 seems like a better amount to show per page. It's also [what BoardGameGeek uses](https://boardgamegeek.com/browse/boardgame), since many people care about "the top 100".

dfabulich:
I cranked it down to 100. Feels pretty good, but I still think I don't want to try it until we have a CDN.

salty-horse:
`gamesearchpopup.php` uses the search API, and will probably need a limit parameter to not show too many entries.

dfabulich:
IFDB is running behind Cloudflare now; I think this is ready to merge.

> `gamesearchpopup.php` uses the search API, and will probably need a limit parameter to not show too many entries.

I think 100 results is basically OK there, too. To test it, I went to a competition page, edited it, clicked "Add Game" and searched for "the".

ifdb.org currently shows 20, which is enough to need to scroll a little. Scrolling a bit more feels fine to me.

alice-blue:
I just tried this on Codespaces. When I click "Browse," The first thing I see is "see all results" block. I guess that is an unfortunate side effect of showing the top 100 games. And the cover art doesn't load, but maybe that's because I'm in Codespaces? I appreciate being able to scroll through the polls faster, though.
screenshot browse limit 100

salty-horse:
That block needs dark mode fixes.

dfabulich:
I've fixed the banner color in dark mode in https://github.com/iftechfoundation/ifdb/pull/943

That banner is due to the "Bury Game" feature. The premise behind it is to restrict the visibility of games where the author has violated our code of conduct (especially by attempting to organize vote manipulation). https://intfiction.org/t/49802

When we were only showing 20 results per page, the first Browse page didn't show the "Some results were hidden" banner. When showing 500 results per page, we did. In the current latest DB archive from Sep 1, the top 100 results include Necromancer at #64, so the banner is visible there.

I think it's a "bad look" to show the banner on the first Browse page.

Luckily, in the current latest version of the DB, Necromancer is now ranked #103, which bumps it off the first page, so the banner won't show.

In a previous iteration of this PR, I just blocked the "some results were hidden" banner on the browse page entirely, but I yanked that out of this PR when I realized that Necromancer had fallen out of the top 100.

alice-blue:
Looks good to me. I can see the cover art now. (The banner is still light colored, but I know you fixed that in a different commit.)

PR #930, opened 2024-10-26
Support JSON output in search page

salty-horse:
* Implemented using "streaming" like the current XML output, so it writes records line by line, instead of serializing a whole JSON object from memory.
* Tested with latin1 and other encodings. To [test latin1](http://localhost:8080/search?searchfor=author%3ALinus+%C5kesson&json), see "Linus Åkesson" author of "The Impossible Bottle" which appears when browsing the top games. To test Unicode, search for [`author:taha`](http://localhost:8080/search?searchfor=author%3Ataha&searchgo=Search+Games&json) Strings with `&#xXXXX;` HTML escape sequences (straight from the database) are converted to unicode.
* yes/no values in XML were turned into booleans for JSON.
* The new `serialize_xml()` can be used in the RSS code as a future cleanup process.
* Most of the boolean values are redundant. You can deduce "hasCoverArt" from the existence of the "coverArtLink" field - can it be removed, maybe just for the new JSON API?
* There's some odd handling of some fields, like "badge" for members. It makes more sense to not return a field when it doesn't exist, rather than an empty element (or empty string in JSON), but I kept it as is.
* What's up with the `tag` search? It's not properly implemented for XML, and returning an error. Should it be removed?
* The [documentation](https://ifdb.org/api/search) needs to be updated.
* I think I used this JSON API in three places - are there more? `xmlSend` can't be used because it's adding the `xml` parameter (which does nothing since `json` in the URL takes precedence, but still...). Maybe it's time for a new modern "send" function based on `fetch` which doesn't try to load ActiveX objects?

### DIFF
--- a/www/gamesearchpopup.php
+++ b/www/gamesearchpopup.php
@@ -62,11 +62,11 @@ function openGameSearchPopup(ele, doneFunc, defaultTitle, openerBtn)
     fld.focus();
     fld.value = defaultTitle;
 }
-function gameSearchPopupKey(event)
+async function gameSearchPopupKey(event)
 {
     var ch = (window.event || event.keyCode ? event.keyCode : event.which);
     if (ch == 13 || ch == 10) {
-        gameSearchPopupGo();
+        await gameSearchPopupGo();
         return false;
     }
     if (ch == 27) {
@@ -81,7 +81,7 @@ function gameSearchPopupClose()
     if (gameSearchPopupOpener)
         setTimeout(function() { gameSearchPopupOpener.focus(); }, 1);
 }
-function gameSearchPopupGo()
+async function gameSearchPopupGo()
 {
     document.getElementById("gameSearchPopupStep2").style.display = "none";
     var txt = document.getElementById("gameSearchPopupSearchBox").value;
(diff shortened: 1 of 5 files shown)


salty-horse:
I forgot about the enum on the same page. I'll add it later today.

salty-horse:
Sorry, I forgot to apply the fixup

salty-horse:
I have a draft of a `jsonSend()` [here](https://github.com/salty-horse/ifdb/compare/json_send?expand=1) - I can try to use it instead of the explicit fetch calls.

salty-horse:
I have two more PRs prepared once this is merged:
1.  `gametags` and `taggame` JSON support (in addition to the existing XML API.)
2. Usage of JSON `taggame` and `taggamedelete` in the website (waiting on #955)

salty-horse:
@dfabulich I see you forced push the branch 1 minute after me! Should I re-push it?

salty-horse:
@dfabulich Can the tag search code I mentioned above be removed?

salty-horse:
Sorry, I left some `error_log`s in. Will remove.

salty-horse:
Forgot to ask this:

I'm currently turning the `&#xXXXX;` to utf-8, but only in the JSON output - should this also be done for the XML output? I think it's nicer that way.

www/gamesearchpopup.php
```diff
@@ -204,17 +206,17 @@ function gameSearchPopupDiv()
...
-                var result = gameSearchPopupKey(event);
+            gameSearchPopupSearchBox.addEventListener('keydown', async function (event) {
+                var result = await gameSearchPopupKey(event);
                 if (result === false) event.preventDefault();
             });
-            gameSearchPopupGoBtn.addEventListener('click', function (event) {
+            gameSearchPopupGoBtn.addEventListener('click', async function (event) {
                 event.preventDefault();
-                gameSearchPopupGo();
+                await gameSearchPopupGo();
```
dfabulich, line 219: This code is _weird_, and I think we should probably clean it up.

In order for `event.preventDefault()` to work, it needs to happen _before_ we `await` a promise. So, the `preventDefault` code here doesn't and can't work.

But, luckily for us, it doesn't _need_ to work! The key handlers are designed to `return false` and thereby `preventDefault` when the user presses the Enter key. (Or the `ch == 10` line feed key?? I don't think any keyboard in modern use has a key that can generate that `charCode`.)

The Enter key would normally submit the form, but on the `editgame` page, we don't want to submit the page's `<form>`.

Luckily, it _doesn't_ submit the form. If you remove the `keypress` and `keydown` listeners entirely, and press Enter, it fires a `click` event on `gameSearchPopupGoBtn` (because it's an `<input type=submit>`), and _that_ prevents default right away.

So, the only purpose of the key handlers is to handle the Escape key, which never even fires `keypress` events, only `keydown` and `keyup` events. Those handlers don't need to be `async` at all. (And they should use `event.code` because `keyCode` is deprecated.)

As for the `click` handler, it doesn't need to be `async` either, because there's no code after `gameSearchPopupGo()`, so there's no point in `await`ing the result.

www/profilelink.php
```diff
@@ -290,13 +294,13 @@ function profileLinkDiv()
...
-            var result = aplPopupKey(event);
+        aplSearchBox.addEventListener('keypress', async function(event) {
+            var result = await aplPopupKey(event);
             if (result === false) event.preventDefault();
         })
-        aplSearchGo.addEventListener('click', function (event) {
+        aplSearchGo.addEventListener('click', async function (event) {
             event.preventDefault();
-            aplSearch();
+            await aplSearch();
```
dfabulich, line 303: We should also clean up these event handlers IMO

www/gamesearchpopup.php
salty-horse: I added the async stuff before I understood that try{} blocks allow use of await.

You're welcome to push more commits to fix these. :)

PR #933, opened 2024-10-28
Prevent cover art from sending a cookie

### LINKED ISSUE #932: Cover art not getting cached in Cloudflare
I think it's refusing to cache because it's passing a `Set-Cookie` header.

```
% curl -o /dev/null -v "https://ifdb.org/viewgame?id=aearuuxv83plclpl&coverart&thumbnail=160x160&version=48"
< date: Mon, 28 Oct 2024 16:35:18 GMT
< content-type: image/png
< strict-transport-security: max-age=63072000; includeSubDomains; preload
< cache-control: public, max-age=31536000, immutable
< set-cookie: PHPSESSID=1dbe71vm40f9jme88b07s5qnkh; path=/; secure; HttpOnly; SameSite=Lax
< content-security-policy: default-src 'self' ifdb.org www.google.com 'nonce-JyZnGxgQ'; script-src 'self' ifdb.org www.google.com 'nonce-JyZnGxgQ'; style-src 'self' ifdb.org 'nonce-JyZnGxgQ'; frame-ancestors 'self';
< speculation-rules: "/speculation-rules"
< cf-cache-status: DYNAMIC
< report-to: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v4?s=OoM6gttgteWaEyRye8bqLiwvCCijMIK9DAfvVJ4O7srnAyhN%2FPqhHJn8v8EPNxDAWUQiPI0t7paTZerpmSqzvg%2BJKYJ%2FtYaVAOG%2B3R%2BtcmG%2FDa%2FejWSYEaIA3Q%3D%3D"}],"group":"cf-nel","max_age":604800}
< nel: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
< server: cloudflare
< cf-ray: 8d9c5698ee5afafc-SJC
< alt-svc: h3=":443"; ma=86400
< server-timing: cfL4;desc="?proto=TCP&rtt=15240&sent=7&recv=10&lost=0&retrans=0&sent_bytes=2880&recv_bytes=607&delivery_rate=150028&cwnd=251&unsent_bytes=0&cid=5ac38e4c7b776e62&ts=133&x=0"
```

dfabulich:
Fixes #932.

When sending cover art, we were starting a session, which adds a `Set-Cookie` header. Cloudflare refuses to cache content with a `Set-Cookie` header, so the cover art wasn't getting cached in Cloudflare.

This required refactoring methods out of `util.php`, which has the side effect of starting a session. `dbconnect.php` no longer references `util.php` (breaking a circular dependency), and the new `image-util.php` contains methods factored out of `util.php`, but `image-util.php` only references `dbconnect.php`.

Unfortunately, it's not really possible to test this locally, because the local dev environment doesn't have an image DB. (I was able to test it locally using the full backup of IFDB, which includes images.)

I was a bit worried about removing the side-effect reference to `util.php` from `dbconnect.php`, so I grepped for files that depend on `dbconnect.php` and not `util.php`. Only four files matched (plus the new `image-util.php`):

* `captcha.php`: Only used from other pages that do include `util.php`. CAPTCHAs still work
* `resetpsw` / `changepsw`: Includes `pagetpl.php` which includes `util.php`. Resetting password still works
* `starctl.php`: I don't think this should include anything at all; I'll file a separate PR for that. But anyway, the star control still works on `viewgame`.


salty-horse (approved):

salty-horse:
The coverart page tries to use `PRODUCTION_SERVER_NAME` which is defined in `session-start.php`:

```php
if (localImageCredentials() === null) {
    header("HTTP/1.1 302 Found");
    header("Location: https://" . PRODUCTION_SERVER_NAME . $_SERVER['REQUEST_URI']);
    exit();
}
```

salty-horse:
@dfabulich  I think this is the reason I don't see images in the local environment. ⬆️

dfabulich:
That's fixed in https://github.com/iftechfoundation/ifdb/commit/d09cc3442ffb4de603b17b324d67ab7edf447b17

Issue #669, opened 2021-03-16
Add a user preference for dark mode


dfabulich:
Apropos iftechfoundation/ifdb-suggestion-tracker#122, in iftechfoundation/ifdb#74 we've added autodetected dark mode.

But now there's no way to opt _out_ of dark mode if your browser is set to "prefer" dark mode except to use a custom stylesheet.

Furthermore, due to the way I've implemented dark mode, that would actually be a pretty tricky stylesheet to write; you'd have to go through and manually undo all of the changes here https://github.com/iftechfoundation/ifdb/blob/main/www/ifdb.css#L2104

Therefore, we should have a user preference to opt in/out of dark mode.

PR #942, opened 2024-10-29
Allow custom stylesheets to force light/dark mode

### LINKED ISSUE #884: Refactor dark mode to avoid copying and pasting styles
We have a bunch of styles in `ifdb.css` under a `@media (prefers-color-scheme: dark)` media rule.

But many users don't use `prefers-color-scheme`, and instead prefer to manually set dark mode on IFDB. Unfortunately, the only way to do that is to use a custom stylesheet; overwhelmingly, the most popular dark mode stylesheet is Midnight.

That means that when we fix stuff in dark mode, a moderator has to login as MJR and manually edit the Midnight style sheet, copying and pasting the current values from `ifdb.css`.

I can think of a few ways to handle this:

1. We could special case Midnight, and write code to make it automatically use whatever's in the dark mode section of `ifdb.css`
2. We could set a `dark` class on the `<html>` element in JS in the `<head>` if the user prefers dark mode and/or if the user has selected Midnight. Then, `ifdb.css` could stop using a `@media` rule and just check for the presence of the `dark` class, and we could just replace the contents of Midnight with a comment or something.
3. We could add a user preference for dark mode #669 and then switch existing Midnight users to the user preference. (But, is this practically different from the second option, above?)

dfabulich:
Fixes #884, and, kinda, #669 

(Note that testing this requires stopping the Docker instance with `docker compose down`, `./prepare_dev_environment.sh`, and then `docker compose up`.

There's a new setting on custom stylesheets, "Force dark mode," which allows a stylesheet to force IFDB into dark mode or light mode.

When dark mode is forced, the rules that are set to `prefers-color-scheme: dark` are set to `prefers-color-scheme: light, prefers-color-scheme: dark`, i.e. the rule applies no matter what.

When light mode is forced, we remove the rules that are set to `prefers-color-scheme: dark`.

When this merges, I'll login as MJR one last time and set the Midnight CSS to be just `@import('ifdb.css')` with force dark mode. Then, future fixes to `ifdb.css` will automatically apply to Midnight users.

Furthermore, it will then be possible for users to pick a stylesheet that opts out of dark mode, even if their OS is set to prefer dark. I guess I'll create a new "Force Light" stylesheet and mention it in site news or something.

### DIFF
--- a/www/styles
+++ b/www/styles
@@ -30,7 +30,6 @@ if ($setID) {
     if ($userid) {
 
         if ($setID == 'default') {
-            $setVal = null;
             $newDesc = "the IFDB default";
         } else {
             // make sure the style sheet actually exists
@@ -82,15 +81,18 @@ if ($setID) {
     $reqTitle = false;
     $reqDesc = false;
     $reqContents = false;
+    $reqDark = false;
     if ($_SERVER['REQUEST_METHOD'] == 'POST') {
         $reqTitle = get_req_data('title');
         $reqDesc = get_req_data('desc');
         $reqContents = get_req_data('contents');
+        $reqDark = get_req_data('dark');
     }
 
     $dbTitle = false;
     $dbDesc = false;
     $dbContents = false;
+    $dbDark = false;
     if ($editID == 'new') {
         // creating a new one - no need to check for an existing copy
         $authorID = $userid;
@@ -100,7 +102,7 @@ if ($setID) {
         $result = mysqli_execute_query($db,
             "select
                s.stylesheetid, s.userid, count(u.id),
-               s.title, s.`desc`, s.contents
+               s.title, s.`desc`, s.contents, s.dark
              from stylesheets as s
                left join users as u
                  on u.stylesheetid = s.stylesheetid and u.id != ?
@@ -110,7 +112,7 @@ if ($setID) {
         $authorID = false;
         if (mysql_num_rows($result) > 0)
             [$ssid, $authorID, $userCnt,
-                 $dbTitle, $dbDesc, $dbContents] = mysql_fetch_row($result);
+                 $dbTitle, $dbDesc, $dbContents, $dbDark] = mysql_fetch_row($result);
     }
 
     if (!$userid) {
@@ -132,6 +134,7 @@ if ($setID) {
         $formTitle = $reqTitle;
         $formDesc = $reqDesc;
         $formContents = $reqContents;
+        $formDark = $reqDark;
 
         // note if it's a Save operation of some kind
         $isSave = isset($_REQUEST['saveStyle'])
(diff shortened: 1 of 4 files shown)


www/styles
```diff
@@ -30,20 +30,17 @@ if ($setID) {
     if ($userid) {
 
         if ($setID == 'default') {
-            $setVal = 'null';
+            $setVal = null;
```
salty-horse, line 33: unused?

www/util.php
```diff
@@ -541,13 +550,25 @@ function showStars($num)
...
     // adding unused class='star-half-checked' and 'star-unchecked' so users can override them in IFDB custom stylesheets
     if (floor($roundedNum) != $roundedNum) {
-        $result .= "<picture><source srcset='/img/dark-images/star-half-checked.svg' media='(prefers-color-scheme: dark)'>"
-            ."<img height=13 class='star-half-checked' src='/img/star-half-checked.svg'></picture>";
+        if ($ssdarkforce === STYLESHEET_DARKFORCE_DARK) {
+            $result .= "<img height=13 class='star-half-checked' src='/img/dark-images/star-half-checked.svg'>";
+        } else if ($ssdarkforce === STYLESHEET_DARKFORCE_LIGHT) {
+            $result .= "<img height=13 class='star-half-checked' src='/img/star-half-checked.svg'>";
+        } else {
+            $result .= "<picture><source srcset='/img/dark-images/star-half-checked.svg' media='(prefers-color-scheme: dark)'>"
```
salty-horse, line 558: can the `media` be on the parent `span` instead of each `source`?

www/styles
dfabulich: fixed

www/util.php
dfabulich: No, `media` is a feature of the `<source>` element. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source

www/styles
```diff
@@ -185,11 +179,10 @@ if ($setID) {
...
                         // add the row
                         $progress = "SIN003";
-                        $result = mysql_query(
+                        $result = mysqli_execute_query($db,
                             "insert into stylesheets
-                             (title, userid, `desc`, contents, created)
-                             values ('$qTitle', '$userid', '$qDesc',
-                                     '$qContents', now())", $db);
+                             (title, userid, `desc`, contents, dark, created)
+                             values (?, ?, ?, ?, now())", [$formTitle, $userid, $formDesc, $formContents, $formDark]);
```
salty-horse, line 185: Missing question mark! 4 instead of 5

salty-horse: Fixed in 62a048dcff1835a30d1e06ed3d0097b32925f76c

PR #954, opened 2024-11-01
Do not leak unwishlist info for games

salty-horse:
The "unwishlist" count was previously removed from the game page, but it was still returned in the API response.

If this information shouldn't be available, it should also be scrubbed from the database dump.

### DIFF
--- a/www/setunwishlist
+++ b/www/setunwishlist
@@ -88,10 +88,9 @@ if ($result) {
     $result = mysql_query(
         "select count(userid) from unwishlists
          where gameid='$qgameid'", $db);
-    list($newCnt) = mysql_fetch_row($result);
 
     // send the success reply
-    sendResponse("Saved", false, "Your change has been recorded.", $newCnt);
+    sendResponse("Saved", false, "Your change has been recorded.");
 }
 else
     sendResponse("Not Saved", "An error occurred updating the database "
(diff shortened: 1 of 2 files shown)


dfabulich:
That data is not private AFAIK. (But maybe it ought to be?)

https://ifdb.org/whoselist?game=uaxh30z0kh7n8nt3&type=unwishlist

alice-blue:
On the settings page, there is a "Public Play Lists" section with checkboxes so that people can decide whether to make their play list / wish list / not interested list public or not

dfabulich:
Oh, yeah, I forgot about that! It looks like they aren't getting scrubbed from the DB properly. 😖 There are 18K unwishlists in the production DB, and 13K unwishlists in the backup.

dfabulich:
As for this PR, I think it makes sense to show the count of public unwishlists in the API. I don't know if it's actually _doing_ that or not…?

salty-horse:
It's not returned in the public API, but it's available in the [whoselist](https://ifdb.org/whoselist?game=aearuuxv83plclpl&type=unwishlist) page, that is not linked from anywhere.

I'm OK with leaving this as-is (I think youtube's video pages still tell you how many thumbs down a video has, even though it's not displayed), but I don't like `setunwishlist` counting and returning something that's unused. Is it ok to merge if I reword the commit message?

dfabulich:
Yeah, that should be fine

salty-horse:
Committed as 7a9e775be6991bcadffe2e5837c2cf306cd6d113

PR #955, opened 2024-11-01
Switch viewgame update APIs to JSON

salty-horse:
This updates `reviewvote`, `setrating`, `setplayed`, `setwishlist`, `setunwishlist` to JSON and forcing POST requests.

It also applies the change from #954 to not leak the unwishlist count.

The new `jsonSend` currently only supports POST requests. While `xmlSend` is used in a few other places for `GET` requests, I think it can be replaced with a simpler `jsonGet`.

### DIFF
--- a/www/ifdbutil.js
+++ b/www/ifdbutil.js
@@ -229,3 +229,48 @@ function forceDarkMode(force) {
         parseSheet(sheet);
     }
 }
+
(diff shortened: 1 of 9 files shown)


salty-horse:
@dfabulich we forced-push together again :D

PR #959, opened 2024-11-02
Streamline competitions and lists; restore `/allnew`

dfabulich:
The `/allnew` page was designed to be the extended version of "New on IFDB," and it's designed to contain a mix of items, so each item begins with a little preamble:

* A new game listing for:
* A new Recommended List by
* A new competition page:

But that preamble is useless in the home page sections. So, in this PR, which looks big, but isn't really _that_ big, I'm copying and pasting the rendering code from `newitems.php` into the home page (refactoring out into new components in each step), and then streamlining the display of components that I didn't already streamline.

Then, I'm restoring the display on `newitems.php`, so `/allnew` looks pretty much the way it used to look.

### DIFF
--- a/www/components/competitions.php
+++ b/www/components/competitions.php
@@ -9,7 +9,65 @@
 <?php
 
 // get the latest competitions and competition news
-showNewItems($db, 0, 4, false, false, false, NEWITEMS_COMPS | NEWITEMS_COMPNEWS);
+$items = getNewItems($db, 7, NEWITEMS_COMPS | NEWITEMS_COMPNEWS);
+
+for ($idx = 0 ; $idx <= 7; $idx++)
+{
+    // get this item
+    list($pick, $rawDate, $row) = $items[$idx];
+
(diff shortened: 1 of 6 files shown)


salty-horse:
PHP warnings:
```
Constant ENABLE_IMAGES already defined in /var/www/html/components/games.php on line 17
Constant ENABLE_IMAGES already defined in /var/www/html/components/reviews.php on line 3
Undefined index: flags in /var/www/html/components/reviews.php on line 18
Undefined index: flags in /var/www/html/components/reviews.php on line 18
Undefined index: flags in /var/www/html/components/reviews.php on line 18
Undefined index: flags in /var/www/html/components/reviews.php on line 18
Undefined index: flags in /var/www/html/components/reviews.php on line 18
Undefined index: flags in /var/www/html/components/reviews.php on line 18
```

I also prefer using `[]` instead of `list()` in new code.

dfabulich:
Fixed!

PR #966, opened 2024-11-02
Use CSS to style checkboxes in game page

salty-horse:
1. Removed apparently-unused `radioBtnWrite` function, and redundant function call chain.
2. Switched checkboxes styling to use CSS instead of Javascript.

Introduced possible issue:

Previously, without Javascript, you couldn't toggle the checkboxes, since they were images. Now, you can check/uncheck them. If this is a problem, we can load the page with those checkboxes disabled, and enable them using Javascript.

In the future it would be nice to have these checkboxes work without Javascript, together with the star ratings.


salty-horse:
Fixed the tag editor, both for editing and deletion. The deletion still uses checkboxes, but maybe they should be buttons. If they are checkboxes, perhaps you should be able to check which tags you want to delete, and only have it be confirmed when click the save button. ATM, the tags go away immediately, so you can't "check your work" before saving changes.

BTW, when I tried to use it with the keyboard, I couldn't focus on any element inside the tag editor popup.

Can you confirm that it's not accessible, and file a new issue if that's the case?

salty-horse:
Should I remove `ckboxCheck` and `ckboxIsChecked`? I think `ckboxGetObj(id).checked` is just as readable.

dfabulich:
This one is conflicted now that the JSON PR has merged

salty-horse:
Fixed conflict.

salty-horse:
I fixed the issues I had with accessibility. Now it's a regular checkbox, and the updating events happen on the `change` event instead of on `click`. Since it's an image, there's no browser-native outline, so I added a thin `teal` outline, which is darker than the "IFDB blue", but looks better.

Used [this guide](https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/). I think it would be nice to switch to SVG checkboxes to match the star rating control.

PR #982, opened 2024-11-03
Home: Make scroll links stick to the top of the page

dfabulich:
I moved the scroll links to the top of the page, above "Welcome to IFDB," and I made them `position: sticky` and styled like a headline. This would have broken scrolling, but I added `scroll-margin-top` on headlines, to leave room for the wrapped links.

image

### DIFF
--- a/www/home
+++ b/www/home
@@ -34,8 +34,23 @@ if ($debugflag) echo "debug mode enabled...<br>";
 
 <?php pageHeader("The Interactive Fiction Database - IF and Text Adventures")?>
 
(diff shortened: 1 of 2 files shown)


salty-horse:
I think the toolbar is too thin, at least on Desktop. If there are links peeking just below the toolbar, it's possible to misclick.

dfabulich:
Added padding, variable scroll-margin-top.

salty-horse:
Better! Personally, I wish the text was vertically centered. There's more "padding" at the top than the bottom.

PR #991, opened 2024-11-04
Larger headers

### LINKED ISSUE #981: Make section headers bigger on the home page
Repeated calls for this on https://intfiction.org/t/an-idea-split-game-and-review-feeds-on-ifdb/66710

dfabulich:
Fixes #981

To get this to work, I had to move links out of the "Welcome to IFDB" section and into "New on IFDB" to prevent wrapping on mobile, and I had to remove "..." from "IFDB Recommends"

# Before

image

# After

image

### DIFF
--- a/www/components/ifdb-recommends.php
+++ b/www/components/ifdb-recommends.php
@@ -365,7 +365,7 @@ if (count($recs) >= 2) {
     $_SESSION['ifdb_recommendations_source'] = $recsrc;
 
     // start the section
-    echo "<div class='headline' id='ifdb-recommends'><h1 class='unset'>IFDB Recommends...</h1>"
+    echo "<div class='headline' id='ifdb-recommends'><h1 class='unset'>IFDB Recommends</h1>"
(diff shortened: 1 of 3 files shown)


salty-horse:
On my local machine, this makes the "New on IFDB" header appears, and is empty.

BTW, unrelated to this, the background of the headers very faint, and I need to concentrate to see its boundaries over the white background.

dfabulich:
Having the section appear blank is kinda by design.

On my first attempt, making "[-] Welcome to IFDB!" larger, alongside the "All Updates | Site News" links, caused the header to wrap to two lines in an ugly way.

The only way to make the header bigger was to move those links somewhere else. But where? I figured I'd move them to the "New on IFDB" section.

But what if there's no site news, competition, poll, or list within the last three days? Where would we put the links then?

And the only reasonable place I could think of was to keep them in the "New on IFDB" section, but leave the section present but empty when there's nothing new.

WDYT? Is there a better thing to do here?

salty-horse:
One solution is to always show at least one news item, even if it's old.

dfabulich:
I like it! Done.

Issue #859, opened 2021-01-26
Use UTF-8


dfabulich:
IFDB stores all of its data in a MySQL `latin1` database and jumps through a bunch of hoops to prevent itself from getting accidentally transmogrified into UTF-8.

When I was porting from MySQL 5 to MariaDB 10, I accidentally restored the database as UTF-8, and so I began the process of trying to adopt UTF-8. I found it didn't seem to be _that_ much work to just use UTF-8. It was more a matter of deleting code than adding it, and making sure nothing got missed.

It would need a pretty serious test plan to make sure nothing got missed, but it seems doable to me.

salty-horse:
If the codebase is prepared in advance to only work with utf-8, is it it just a matter of shutting down the website for a bit and converting the database?

dfabulich:
While working on #995, I learned something that surprised me: **our database is in Windows-1252, not ISO-8859-1.**

[Windows-1252 is basically a superset of 8859-1.](https://stackoverflow.com/a/75789521/54829) There are a couple dozen characters that are defined as "control characters" in ISO-8859-1 and in UTF-8, but are defined as visible characters in 1252, and the smart apostrophe `\x92` is one of them.

`\x92` is all over our database, in reviews, etc. In ISO-8859-1 it's a non-renderable control character. It doesn't even display properly as an XML entity.

PR #995, opened 2024-11-04
Fix RSS encoding

dfabulich:
RSS was attempting to convert from ISO-8859-1 to UTF-8 but it wasn't working correctly; I'm not entirely sure why.

But there's a much easier way: converting all non-ASCII characters (e.g. smart quotes) into XML entities. It means that we never have to worry about charsets, automatic charset translation, etc. etc.

So, that's what this PR does.

### DIFF
--- a/www/allnew-rss
+++ b/www/allnew-rss
@@ -16,7 +16,7 @@ $numItems = 25;
 iconv_set_encoding("output_encoding", "UTF-8");
 
 // send the RSS content-type header
-header("Content-Type: application/rss+xml");
+header("Content-Type: application/rss+xml; charset=utf-8");
 header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
 header("Cache-Control: no-store, no-cache, must-revalidate");
 header("Cache-Control: post-check=0, pre-check=0", false);
--- a/www/commentlog
+++ b/www/commentlog
@@ -137,7 +137,7 @@ if ($pg == 'all') {
 if ($rss) {
 
     // send the RSS content-type header
-    header("Content-Type: application/rss+xml");
+    header("Content-Type: application/rss+xml; charset=utf-8");
     header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
     header("Cache-Control: no-store, no-cache, must-revalidate");
     header("Cache-Control: post-check=0, pre-check=0", false);
--- a/www/news
+++ b/www/news
@@ -27,7 +27,7 @@ if (isset($_REQUEST['rss'])) {
     iconv_set_encoding("output_encoding", "UTF-8");
 
     // send the RSS content-type header
-    header("Content-Type: application/rss+xml");
+    header("Content-Type: application/rss+xml; charset=utf-8");
     header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
     header("Cache-Control: no-store, no-cache, must-revalidate");
     header("Cache-Control: post-check=0, pre-check=0", false);
--- a/www/poll
+++ b/www/poll
@@ -996,7 +996,7 @@ administrators can see who voted for what.)
         iconv_set_encoding("output_encoding", "UTF-8");
 
         // send the RSS content header
-        header("Content-Type: application/rss+xml");
+        header("Content-Type: application/rss+xml; charset=utf-8");
         header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
         header("Cache-Control: no-store, no-cache, must-revalidate");
         header("Cache-Control: post-check=0, pre-check=0", false);
--- a/www/putific
+++ b/www/putific
@@ -16,7 +16,7 @@ $db = dbConnect();
 //
 function replyXml($body)
 {
-    header("Content-type: text/xml");
+    header("Content-type: text/xml; charset=utf-8");
     header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
     header("Cache-Control: no-store, no-cache, must-revalidate");
     header("Cache-Control: post-check=0, pre-check=0", false);
--- a/www/reviewcommentlog
+++ b/www/reviewcommentlog
@@ -127,7 +127,7 @@ list($rowcnt) = mysql_fetch_row($result2);
 if ($rss) {
 
     // send the RSS content-type header
-    header("Content-Type: application/rss+xml");
+    header("Content-Type: application/rss+xml; charset=utf-8");
     header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
     header("Cache-Control: no-store, no-cache, must-revalidate");
     header("Cache-Control: post-check=0, pre-check=0", false);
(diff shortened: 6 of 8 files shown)


salty-horse:
Is there a specific example that is currently breaking?

I use `rss_encode` in the API output so I'd like to test it.

dfabulich:
As of right now, <https://ifdb.org/allnew-rss> "Kiss of Beth"

```
<description>I canÂ’t decide
```

You can repro it locally by posting a review with a smart quote. Notably, Chinese characters are converted into XML entities `&#xxxx;` so encoding doesn't matter there.

In the past I've enjoyed publishing XML as ASCII, converting _all_ non-ASCII characters (e.g. smart quotes) into XML entities. It means that I never have to worry about charsets, automatic charset translation, etc. etc.

But I couldn't find a straightforward way to do that in PHP. I think it would be great if `rss_encode` did that.

dfabulich:
I googled it again and found the way.

salty-horse:
I need to investigate. I don't understand why this is needed (I thought those verbose entities only belong in the DB which doesn't support unicode), and this breaks my JSON output where I make use of `rss_encode` and `html_entity_decode` to go from latin1 and entities to raw utf-8.

salty-horse:
`iconv("windows-1252", "UTF-8", $str);` is solving the smart quote issue (which I've looked up before [[1]](https://stackoverflow.com/questions/51517153/why-is-u0092-symbol-not-visible-in-console-debug-window-but-its-counterpart) [[2]](https://stackoverflow.com/questions/51517153/why-is-u0092-symbol-not-visible-in-console-debug-window-but-its-counterpart))

With this, I have no problem viewing the smart quote in allnew-rss, or in my xml or json views of the search page. I don't understand why you want to convert everything to entities. In UTF-8 XML, quoting is only needed for the 4 special characters that `htmlspecialcharx` handles.

salty-horse:
Forgot to mention that even with the existing behavior, I don't see the `Â’t` you describe.

The U+0092 appears as a raw 0092 character for me in that specific review. Here it is in RSS since allnew-rss changes rapidly:
`curl 'https://ifdb.org/viewgame?id=lruuhm7se68ugnz6&reviews&rss' | gview -`

dfabulich:
> I don't understand why you want to convert everything to entities.

> I don't see the Â’t you describe.

That's exactly the point of using entities. Diagnosing and reproducing character encoding issues is an enormous pain, because whenever you try to "view" the document to see what encoding it's in, the viewer might try to decode the text, and it could make confusing assumptions as it does.

With ASCII + entities, you and I can objectively agree that the text says `can&#146;t` (and, if it doesn't, or if it's using the wrong entity, we can fix it, and easily see that it's fixed).

(I recommend sending ASCII with entities for JSON, too. Not XML entities, of course, but `\u####` entities, for everything but 7-bit ASCII. This ensures that everything will display and decode the JSON in exactly the same way.)

As for reproducing the issue, the RSS feed validator can reproduce it. https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fifdb.org%2Fallnew-rss

> image

I don't have `gview` on my machine, but I hope that's sufficient to persuade you that the issue is real _and_ hard to diagnose.

(I noticed the problem because the IFDB RSS feed has invalid characters in Feedly, my preferred feed reader.)

FWIW, I'm viewing the feed in Google Chrome, with no special extensions, to see the `canÂ’t`.

> `iconv("windows-1252", "UTF-8", $str);` is solving the smart quote issue

I kinda agree, but in order to show `can’t` in Google Chrome, I also have to add `charset=utf-8` to the `Content-Type` header, which means updating all the places we did that. (The previous version of this branch did that.)

But that creeped me out, because I don't understand why `iconv("windows-1252", "UTF-8", $str);` seems to help. Our database is supposedly using `latin1` aka ISO-8859-1.

```
CREATE DATABASE IF NOT EXISTS ifdb CHARACTER SET latin1 COLLATE latin1_german2_ci;
```

`pagetpl.php` says 8859-1. 8859-1 appears all over the IFDB code base. If we add `windows-1252` here to `rss_encode`, it will be the first time the string `1252` appears in IFDB's code. Why would `windows-1252` be needed _here_ and _only here_???

The fix I have here makes sense to me. It's converting ISO-8859-1 into XML entities that are easy to verify and definitely work. We don't have to send any out-of-band signals to Chrome or any feed reader to ensure they interpret our XML entities exactly the way we intended.

salty-horse:
`gview` is just `vim` in read-only mode (`view`) with a GUI :)

1. The validator doesn't complain about the output when using iconv with `windows-1252` when I paste it, and "interoperability with the widest range of feed readers". Can you maybe push this change to dev, so we can check if it complains about the lack of encoding information in the headers and the xml document?
2. Where is Chrome giving you trouble? When viewing the RSS? For me in Chrome is just shows that page monospaced text, with quotes and smart quotes appearing correctly. The encoding isn't specified, but it's assuming UTF-8 correctly. This is using version 129.0.6668.100 on Linux.
3. The difference between the ISO-8859-1 and windows-1252 encodings that's relevant here is that the ISO standard defines the small range 0x80…0x9F used by smart quotes as control characters, and not printable characters. And a few other characters are also unsupported, like the euro sign € [More here](https://stackoverflow.com/a/48922369). Try adding a review with these symbols locally and then view it in the allnew-rss. The `windows-1252` thing fixes the RSS display, and even without it, everything goes fine from the browser to the server to the database and back.

I understand your concern about compatibility, but I'm still not convinced more is needed beyond the change I suggested. If the validator is fine with it, what's the worry?

After all the JSON business, I'm willing to tackle #859, and convert the database to UTF-8 (including converting the Unicode-as-HTML=entities to proper UTF-8).

dfabulich:
> 2. Where is Chrome giving you trouble? When viewing the RSS? For me in Chrome is just shows that page monospaced text, with quotes and smart quotes appearing correctly. The encoding isn't specified, but it's assuming UTF-8 correctly. This is using version 129.0.6668.100 on Linux.

When I go to https://ifdb.org/allnew-rss in Chrome 130.0.6723.92 for macOS, it's showing `ItÂ’s a bitsy game`. When I open Chrome Dev Tools while viewing that URL and run `document.characterSet` it prints `windows-1252`, which is a wrong guess.

dfabulich:
I have some leftover UTF-8 code from when I first tried it. Let me dig it up.

salty-horse:
This Chrome behavior is after applying the windows-1252 change?

dfabulich:
I've reverted this branch back to my original strategy, which uses `windows-1252`, and just adds `charset` declarations on all of our XML pages. It works fine for me now, but we may or may not ship it, if we just want to rip the bandaid off and use UTF-8 everywhere.

salty-horse:
I doubt that charset header is required, as it's default. Is it just to be on the safe side, or did you see it causing problems anywhere?

I suggest shipping it, as we already know it only affects a small range of characters, for the better.

When I added JSON output, I also rewrote the XML output (to support both having the same input source). I tested both "latin1 range" and higher "unicode range" characters, and everything worked fine without explicitly stating the encoding. UTF-8 is guessed correctly (See "Other" [in this table of how it guesses the encoding](https://www.w3.org/TR/REC-xml/#sec-guessing-no-ext-info))

dfabulich:
I'm gonna fix this "properly" with #998

PR #998, opened 2024-11-05
Adopt UTF-8

### LINKED ISSUE #859: Use UTF-8

### LINKED ISSUE #997: Tags with curly apostrophes and quotes are saved incorrectly
Navigate to a game and add the tag `can’t`, which contains a curly apostrophe.

When you refresh the page, you'll see the tag has been saved as `can&#8217;t`. In fact, if you query the database, you'll find that it was _saved_ that way; this isn't just a display error.

dfabulich:
Fixes #859

You probably want to review this commit by commit, because in a few places it's just a bunch of find-and-replace that can be hard to review.

The first commit runs a schema patch to convert all the tables to UTF-8, and fixes a bunch of central places where we were forcing 8859-1 instead of UTF-8. This covered 95% of what we needed to fix.

The second commit fixes #997, so you can save a tag `can’t` which then gets converted to `can't` with a straight ASCII apostrophe (so users searching for `'can't` with a straight apostrophe will match it)

The rest is just cleanup.

TODO:
* Rip out `encodeURI8859` and replace it with `encodeURIComponent` (this causes a ton of conflicts with open PRs, so I'll hold off for now, but it's a simple regex find-and-replace)
* Rip out `rss_encode` (ditto; lots of conflicts)
* IF Archive scrubber
* After the next public archive gets posted in December, update `create_db.sql` and `patch-schema.sql` to incorporate UTF-8.


salty-horse:
Another optional is converting all the HTML entities in the database to utf-8.

This can be done programmatically by going over all the tables, finding the text-like columns, and scanning rows one by one.

In PHP the function is `html_entity_decode`, and in Python it's `html.unescape`. I don't think node has it built-in.

I can write it if you want.

dfabulich:
I've filed converting HTML entities separately as #1002.

PR #999, opened 2024-11-05
Support JSON in game tags APIs

salty-horse:
This updates `gametags`, `taggame`, `taggamedelete` to use JSON.

`gametags` and `taggame` are documented APIs, so they still support XML as well as JSON. However, in JSON-mode, `taggame` now requires POST, and has a different interface.

Both `taggame` and `taggamedelete` now have accept their input as JSON, and have JSON output. Since they share some code, I moved it to a different function. I think it might be nicer to keep both "tag" and "delete" functionality in the endpoint. What do you think?

The `gametags` response query was really weird, so I updated it to use a subselect - I checked it closely to confirm it's doing the same thing, but please have a look too.

After this, I'll submit updates to `ifwiki-check` and `capreq`, and remove `xmlreq.js`.

### DIFF
--- a/www/gametags
+++ b/www/gametags
@@ -6,15 +6,37 @@ include_once "login-check.php";
 
 function sendResponse($status, $errmsg, $tagInfo)
 {
+    global $json;
+
     header("HTTP/1.1 $status");
-    header("Content-Type: text/xml");
     header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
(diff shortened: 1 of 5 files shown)


salty-horse:
I tried to merge taggame and taggamedelete into one file, and I think it's better as-is.

salty-horse:
Resolved conflicts after UTF-8 changes, but there's an error deleting UTF-8 tags. Will check later.

salty-horse:
Fixed.

PR #1001, opened 2024-11-05
Add JSON output to viewgame

salty-horse:
Returns JSON when the query has `ifiction&json`. Maybe `json` can be enough since it's not really the standard ifiction format.

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -498,55 +498,70 @@ if (!$errMsg && isset($_REQUEST['rss'])) {
     exit();
 }
 
-// check for an iFiction request
-function xmlField($fld, $val)
+$json = isset($_REQUEST['json']);
+if (isset($_REQUEST['ifiction']) || $json)
 {
-    if ($val)
-        echo "<$fld>" . rss_encode(htmlspecialcharx($val)) . "</$fld>";
-}
-function xmlDescField($fld, $val)
-{
-    if ($val)
-        echo "<$fld>" . rss_encode(fixDesc($val, FixDescIfic)) . "</$fld>";
-}
-if (isset($_REQUEST['ifiction']))
-{
-    header("Content-Type: text/xml");
-    header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
-    header("Cache-Control: no-store, no-cache, must-revalidate");
-
-    echo "<?xml version=\"1.0\" encoding=\"UTF-8\"?>";
+    if (!$json) {
+        header("Content-Type: text/xml");
+        header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
+        header("Cache-Control: no-store, no-cache, must-revalidate");
+    }
 
     // if an error occurred, show the error
     if ($errMsg) {
-        echo "<viewgame xmlns=\"http://ifdb.org/api/xmlns\">";
-        xmlField("errorCode", $errCode);
-        xmlField("errorMessage", $errMsg);
-        echo "</viewgame>";
+        $error_fields = [
+            'errorCode' => $errCode,
+            'errorMessage' => $errMsg,
+        ];
+        if ($json) {
+            send_json_response($error_fields);
+        } else {
+            echo "<viewgame xmlns=\"http://ifdb.org/api/xmlns\">";
+            echo serialize_xml($error_fields);
+            echo "</viewgame>";
+        }
         exit();
     }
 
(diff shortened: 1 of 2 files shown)


dfabulich:
1. I do think it should be `json` and not `ifiction&json`, because this format isn't iFiction™. It bears some similarity to it, but the format we're using isn't standardized.

    (It's possible in principle to standardize JSON iFiction if you want to submit PRs to https://github.com/iftechfoundation/ifarchive-if-specs and https://github.com/iftechfoundation/babel-tool/ but I'm not sure it's worth anyone's time.)
2. I think this is an opportunity to return full HTML in the `description` field, instead of following the [iFiction rules](https://babel.ifarchive.org/babel.html#description). e.g. Anchorhead's description starts with an italicized passage, which is not permitted in iFiction, but this is JSON, so we can do anything we want.

salty-horse:
Pushed a fix for `&json

Which flags of fixDesc should be used instead? `FixDescSpoiler`, `FixDescRSS`, or None? Looks like they mostly affect how spoiler tags are handled.

dfabulich:
I think None, but we can test it out. We're using None mode here: https://github.com/iftechfoundation/ifdb/blob/main/www/gameinfo.php#L206

And we seem to be using that description with no further fixups on the page.

(FWIW, descriptions aren't allowed to contain spoiler tags; those are only allowed in reviews and comments.)

www/viewgame
```diff
@@ -643,41 +658,62 @@ if (isset($_REQUEST['ifiction']))
...
-    // main end tags
-    echo "</story>"
-        . "</ifindex>";
+    // Convert object from latin1 to UTF-8
+    array_walk_recursive($story_obj, function(&$value, $key) {
+        if (is_string($value)) {
+            $value = rss_encode($value);
+            $value = html_entity_decode($value);
+        }
+    });
```
dfabulich, line 704: Now that we're using UTF-8 in the DB, I think we should fix #1002 first before adding new code like this

www/viewgame
```diff
@@ -593,47 +607,48 @@ if (isset($_REQUEST['ifiction']))
...
+    if ($firstpublished)
+        $bibliographic['firstpublished'] = $firstpublished;
 
-    echo "</bibliographic>"
-        . "<contacts>";
+    if ($genre)
+        $bibliographic['genre'] = $genre;
 
-    xmlField("url", $website);
+    $desc = fixDesc($rawDesc, FixDescIfic);
```
dfabulich, line 638: Let's do None mode and see how it looks

dfabulich (changes requested):

salty-horse:
Done. Tested on Anchorhead.

www/viewgame
salty-horse: I think it's the opposite - `html_entity_decode` can be removed AFTER #1002 is done. JSON strings are expected to be Unicode (or `\uxxxx`, and not `&#xxxx;`. For XML it's the same either way - my browser converted the `&#xxxx;` automatically for display, and now it doesn't have to.

BTW the same bit of code is in `/search`.

salty-horse: However, `rss_encode` shouldn't be in new code, will remove.

dfabulich: Yeah, I misspoke: once #1002 is done, we won't need this recursive walk at all

www/viewgame
```diff
@@ -640,41 +655,61 @@ if (isset($_REQUEST['ifiction']))
...
 
-    // main end tags
-    echo "</story>"
-        . "</ifindex>";
+    // Convert object from latin1 to UTF-8
+    array_walk_recursive($story_obj, function(&$value, $key) {
+        if (is_string($value)) {
+            $value = html_entity_decode($value);
+        }
+    });
```
dfabulich, line 700: As we were discussing earlier, I've merged #1002 and run the admin tool in production (and I filed #1026 to remove special cases for HTML entities).

I think we should remove this `array_walk` code and merge #1026, and then merge this PR.

dfabulich (changes requested):

salty-horse:
I rebased and force-pushed, so the "compare" button on the force-push line will show a lot of the commits that happened in-between, but I can assure you that I just removed the array_walk :)

salty-horse:
I can remove the duplication of the xml headers, and print them once at the top. Sec.

salty-horse:
Ready :)

Issue #606, opened 2021-04-04
Avoid querying IFWiki whenever a game page is loaded


salty-horse:
When a browser loads the game page, it makes an AJAX call to IFDB, which makes an HTTP call to IFWiki, looking for a wiki page for the game in question (based on the first IFDB).

This is done every time a page is loaded, which is unnecessary. As discussed in iftechfoundation/ifdb#123, the result should be cached for a reasonable period of time, and the IFWiki link should be returned as part of the page.

* Should the caching be done in a new column in the database?
* What is a reasonable period of time for which to save positive/negative results?
* Which action should cause the cache to be cleared?

salty-horse:
Here's [an attempt](https://github.com/salty-horse/ifdb/compare/ifwiki_clientside?expand=1) to query IFWiki directly from the browser instead of going through IFDB, so there's no need for the `ifwiki-check` API.

It's giving an error about a missing `Access-Control-Allow-Origin` - Is that missing on IFWiki? Is there away to "allow" it with headers on IFDB?

dfabulich:
#1024 moved the request to IFWiki, but I think we should still not bother pinging IFWiki all the time.

My favorite answer is to have a scheduled job that updates an IFDB cache table of IFWiki links.

alice-blue:
Jonathan said elsewhere:

Here are some examples of what IFDB could do now, by using the IFWiki API instead of relying on those manually created “IFDB…” wiki pages (which were not always created, and won’t be any more). @dfabulich

Get all pages that have either an IFID or TUID
`Query: https://www.ifwiki.org/api.php?action=cargoquery&tables=Games&fields=_pageName=pageName,TUID,IFID&where=TUID__full%3C%3E%22%22%20AND%20IFID__full%3C%3E%22%22&limit=500`

Jonathan also said:

You can use the offset parameter to get beyond the first 500.

PR #1003, opened 2024-11-05
Switch ifwiki-check to communicate in HTTP status codes

salty-horse:
Removes another use of xmlSend.

Ideally, the browser would contact IF Wiki directly, as mentioned in #606.

### DIFF
--- a/www/ifwiki-check
+++ b/www/ifwiki-check
@@ -13,21 +13,11 @@ curl_exec($ch);
 $status_code = curl_getinfo($ch, CURLINFO_HTTP_CODE);
 curl_close($ch);
 
-$exists = ($status_code == 200);
-
+header("HTTP/1.1 $status_code");
 header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
 header("Cache-Control: no-store, no-cache, must-revalidate");
 header("Cache-Control: post-check=0, pre-check=0", false);
 header("Pragma: no-cache");
-header("Content-type: text/xml");
-
-echo "<?xml version=\"1.0\" encoding=\"UTF-8\" ?>"
-   . "<ifwiki-check-result>"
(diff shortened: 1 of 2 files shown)


www/viewgame
```diff
@@ -2846,12 +2846,9 @@ dispTags();
...
                         <script nonce="<?php global $nonce; echo $nonce; ?>">
-                            xmlSend("ifwiki-check?ifid=<?php echo $jifid ?>", null,
-                                function(doc) {
-                                    if (xmlChildText(doc, "exists") == "yes") {
-                                        document.getElementById("ifwiki-link").style.display = "initial";
-                                    }
-                                }, null);
+                            fetch("ifwiki-check?ifid=<?php echo $jifid ?>").then(() => {
+                                document.getElementById("ifwiki-link").style.display = "initial";
+                            });
```
dfabulich, line 2851: This isn't checking the status. <http://localhost:8080/viewgame?id=qvcdjqvl43b0whlx> is showing an IFWiki link, but it shouldn't.

dfabulich (changes requested):

salty-horse:
Fixed!

www/ifwiki-check
```diff
@@ -13,21 +13,11 @@ curl_exec($ch);
 $status_code = curl_getinfo($ch, CURLINFO_HTTP_CODE);
 curl_close($ch);
 
-$exists = ($status_code == 200);
-
+header("HTTP/1.1 $status_code");
```
dfabulich, line 16: I prefer to use [`http_response_code()`](https://www.php.net/manual/en/function.http-response-code.php). (It's fine, either way.)

salty-horse: Please change it, because I won't remember in the future without good examples to look at.

salty-horse: (I copied this from another place in the code)

PR #1004, opened 2024-11-06
Fix series and genre links containing non-ASCII characters

### LINKED ISSUE #508: viewgame links to some series or genre names containing non-ASCII characters don't work
There are two listings for "Cannelé & Nomnom" games: [beta](https://ifdb.org/viewgame?id=px4i00emsjwozza0) and [episode 1](https://ifdb.org/viewgame?id=i57zntqzl6fnkt5v).

They are marked as belonging to a series called _"Cannelé & Nomnom"_.

Clicking on the series link in either of the game pages produces (in Firefox) a _"No results were found"_ page, with the text field displaying _"series:CannelÃ© & Nomnom"_ (that's A-tilde and copyright-sign in place of the expected e-acute).

The search result link in the game page source is `/search?searchfor=series:Cannel%C3%A9+%26+Nomnom` (resolving to [this link](https://ifdb.org/search?searchfor=series:Cannel%C3%A9+%26+Nomnom) as an absolute URL). The %-encoded parts (C3 A9) represent the UTF-8 encoding of U+00E9 LATIN SMALL LETTER E WITH ACUTE. (But when interpreted as ISO-8859-1, they produce the Ã© mojibake seen above.)

If I hand-hack the URL to encode the ISO-8859-1 (latin1) representation of that character (i.e., `series:Cannel%E9+%26+Nomnom`) -- [link](https://ifdb.org/search?searchfor=series:Cannel%E9+%26+Nomnom) -- that works (returns both games from the series, sensible search box contents).

I don't know anything about IFDB's internal character representation (and I note open issue iftechfoundation/ifdb#859), but all the pages I've looked at (/viewgame, /editgame, search result) are served with HTTP header `Content-Type: text/html; charset=ISO-8859-1`, have the same in `<meta http-equiv>`, and contain literal `E9` bytes where "Cannelé" is mentioned. <s>So I guess it's generally Latin-1, and the series name in these games' data is fine, and whatever is transcoding the series name to UTF-8 for the series URL needs squashing.</s>

dfabulich:
Fixes #508

EDIT (Nov 7): This PR is redone from scratch now, in three commits.

The first commit trims our use of `getGameInfo` in `showuser`; we only need `$links`, and now the code shows that.

The second commit is a tiny bit scary, security-wise. `getGameInfo` used to take care of running `htmlspecialcharx` on some but not all strings it returns. Now, it doesn't pre-run `htmlspecialcharx` at all. `showuser` doesn't care, because it was only ever using the links, but now `viewgame` calls `htmlspecialchars` itself right after destructuring the result.

(I tested this by adding `<` to the series, genre, author, and system of <http://localhost:8080/viewgame?id=i57zntqzl6fnkt5v>, viewing it in XML and regular mode.)

In the second commit, we delay `htmlspecialchars` for `$system`, `$genre`, and `$seriesname`, so we can separately `urlencode` them for searching and `htmlspecialchars` encode them for display. This strategy now matches the approach we use for `$author`.

We're also now using `urlencode` and not `urlencodeFromHTML`, which was built on the assumption that the database was storing HTML entities, and that the data would be in ISO-8859-1 (Windows-1252), which it converted using [`utf8_encode`](https://www.php.net/manual/en/function.utf8-encode.php).

> utf8_encode — Converts a string from ISO-8859-1 to UTF-8
> WARNING This function has been DEPRECATED as of PHP 8.2.0. Relying on this function is highly discouraged.

### DIFF
--- a/www/showuser
+++ b/www/showuser
@@ -197,22 +197,8 @@ if ($rss == 'gamenews')
         $items[] = array($moddate, $item);
 
         // Now add news updates for the game itself.
-        list($ifids, $title, $author, $authorExt,
-             $pubYear, $pubFull, $license,
-             $system, $desc, $rawDesc,
-             $hasart, $genre, $seriesname, $seriesnum,
-             $forgiveness, $bafsid, $version,
-             $language, $languageNameOnly,
-             $website, $links,
-             $ratingcountavg, $ratingcounttot, $ratingavg, $memberReviewCnt,
-             $currentUserRating, $currentUserReview,
-             $editedbyid, $editedbyname, $moddate, $moddate2, $pagevsn,
-             $moddatelatest,
-             $historyView,
-             $dlnotes, $extReviews, $extRevDisplayRank,
-             $ratingHisto, $xrefs, $inrefs) =
-                 getGameInfo($db, $gameid, $curuser, false,
-                             $gameErrMsg, $gameErrCode);
+        $links = getGameInfo($db, $gameid, $curuser, false,
+                        $gameErrMsg, $gameErrCode)[20];
 
         // add the RSS items for this game to the master list
         if (!$gameErrMsg) {
--- a/www/viewgame
+++ b/www/viewgame
@@ -388,9 +388,6 @@ if (!$errMsg) {
     $title = htmlspecialchars($title);
     $license = htmlspecialchars($license);
     $language = htmlspecialchars($language);
-    $system = htmlspecialchars($system);
-    $genre = htmlspecialchars($genre);
-    $seriesname = htmlspecialchars($seriesname);
     $seriesnum = htmlspecialchars($seriesnum);
     $forgiveness = htmlspecialchars($forgiveness);
 
@@ -1440,8 +1437,8 @@ if (!isEmpty($seriesname)) {
         echo "Part of ";
     else
         echo "Episode $seriesnum of ";
-    echo "<a href=\"search?searchfor=series:" . urlencodeFromHTML($seriesname)
-        . "\">$seriesname</a><br>";
+    echo "<a href=\"search?searchfor=series:" . urlencode($seriesname)
+        . "\">".htmlspecialchars($seriesname)."</a><br>";
 }
                      ?>
                      <?php if (!isEmpty($pubYear)) echo "$pubYear<br>" ?>
@@ -1449,11 +1446,11 @@ if (!isEmpty($seriesname)) {
 if (!isEmpty($genre)) {
     $genres = explode(",", str_replace("/", ",", $genre));
     echo join(", ", array_map(
-        fn($genre) => "<a href=\"search?searchfor=genre:" . urlencodeFromHTML($genre) . "\">$genre</a>",
+        fn($genre) => "<a href=\"search?searchfor=genre:" . urlencode($genre) . "\">".htmlspecialchars($genre)."</a>",
         $genres)) . "<br>";
 }
                      ?>
-                     <?php if (!isEmpty($system)) echo "<div>$system</div>" ?>
+                     <?php if (!isEmpty($system)) echo "<div>".htmlspecialchars($system)."</div>" ?>
                      <?php
                         $showExternalLinksAnchor = count($links) && ($detailView || $historyView);
                         if (!isEmpty($website) || $showExternalLinksAnchor) {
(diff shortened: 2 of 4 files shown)


salty-horse:
`viewgame` handles authors [differently than the other links](https://github.com/iftechfoundation/ifdb/blob/59fe978b66c3926d9f22a2d406927b6095d472e5/www/viewgame#L1418). Should it be converted to use this function?

```php
// quote the author's name for display and use in the search HREF
$uau = urlencode($au);
$dau = htmlspecialcharx($au);
                                                                         
// show the link
echo "$conj<a href=\"search?searchfor=author%3A$uau\">$dau</a>$proLink";
```

salty-horse:
Is #1027 (about tags with `"`) related?

dfabulich:
I don't think so, but, to prove it, I'll fix it separately.

dfabulich:
OK! This PR is ready for review, I think. I redid it from scratch; see the edited description at the top of the PR thread.

www/showuser
```diff
@@ -213,6 +213,14 @@ if ($rss == 'gamenews')
              $ratingHisto, $xrefs, $inrefs) =
                  getGameInfo($db, $gameid, $curuser, false,
                              $gameErrMsg, $gameErrCode);
+        $title = htmlspecialchars($title);
```
salty-horse, line 216: `showuser` doesn't use most of these, so no need to escape them. Is this for "safety"?

Unused: `$license`, `$language`, `$system`, `$genre`, `$seriesname`, `$seriesnum`, `$forgiveness`.

(In the future I want getGameInfo to return a dict, so the caller can destructure just the fields it needs.)

www/viewgame
```diff
@@ -1432,20 +1437,20 @@ if (!isEmpty($seriesname)) {
         echo "Part of ";
     else
         echo "Episode $seriesnum of ";
-    echo "<a href=\"search?searchfor=series:" . urlencodeFromHTML($seriesname)
-        . "\">$seriesname</a><br>";
+    echo "<a href=\"search?searchfor=series:" . urlencode($seriesname)
```
salty-horse, line 1440: I don't remember how the search "unpacks" this data in reverse. It's not enough to `urlencode` this? It has to also be HTML-escaped?

www/showuser
dfabulich: Yeah, I didn't want to bother to prove that `showuser` definitely _doesn't_ use any of these.

www/viewgame
salty-horse: I'm not sure if you answered exactly what I was asking.

Now it's basically `urlencode(htmlspecialchars($raw_seriesname))` - is that the intention? Just to pass the data through the URL medium, it's enough to use `urlencode()` with `%XX`. Does the other side (search) expect it to also be HTML-escaped with `&XXX;`?

dfabulich: I claim that line 1440 is now calling `urlencode($raw_seriesname)`, and then in the body of the `<a>` tag we're displaying `htmlspecialchars($raw_seriesname)`.

The other side (search) no longer expects or requires the data to be HTML-escaped. Search just pulls out the `$seriesname` raw from the `$_REQUEST` and uses it.

www/showuser
dfabulich: We weren't even using the `$title`! We were only ever using the `$links`. I've updated the PR to first make `showuser` _only_ grab `$links = getGameInfo(...)[20]`, which proves that we don't need to `htmlspecialchars` anything here.

www/viewgame
salty-horse: Oh, sorry, I missed that `$seriesname` was not in the list above. OK :)

www/showuser
```diff
@@ -197,22 +197,8 @@ if ($rss == 'gamenews')
...
-             $currentUserRating, $currentUserReview,
-             $editedbyid, $editedbyname, $moddate, $moddate2, $pagevsn,
-             $moddatelatest,
-             $historyView,
-             $dlnotes, $extReviews, $extRevDisplayRank,
-             $ratingHisto, $xrefs, $inrefs) =
-                 getGameInfo($db, $gameid, $curuser, false,
-                             $gameErrMsg, $gameErrCode);
+        $links = getGameInfo($db, $gameid, $curuser, false,
+                        $gameErrMsg, $gameErrCode)[20];
```
salty-horse, line 201: must not forget to tweak that magic 20 if more return values are added before its position. :(

salty-horse (approved):

www/showuser
dfabulich: There are only two callers for `getGameInfo`, and messing around with argument position is obviously risky, so I'm not too worried.

PR #1009, opened 2024-11-06
Use JSON in captcha code

salty-horse:
Also fixed issues in `jsonSend` error handling.

To test the captcha, you can use [the keys from here](https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do). I suggest including them in the development environment for everyone, if there's such a settings file.

Then you want to remove admin privs from the admin account, and add an email to the user. Captchas are also shown for yourself in `/showuser`, so this will do:
```mysql
update users set privileges='A' where id='0000000000000000';
update users
set email="test@example.com", publicemail="foobar@test.com", emailflags=1
where id='0000000000000000';
```

I also fixed an issue with `Recaptcha.reload()` not being defined - I think it's now `reset()`, and the canonical name of the object is now `grecaptcha`.

Finally, I deleted `submitCaptchaForm()`. I'm not sure what's it for. It shares code with `captchaSolved`. Both are in the form, but `submitCaptchaForm()` is in a "submit" event that's never triggered, because there's no submit button. Am I missing something? Maybe it's for an older version of reCAPTCHA?

```html
<form name='captchaAjaxForm'><script nonce='XXw5d8kB'>
document.currentScript.parentElement.addEventListener('submit', function (event) {
var result = (function(){ submitCaptchaForm();return false; }).apply(event.target);
if (result === false) event.preventDefault();
});
</script>
<div id='captchaFormCont'></div><div><span id='captchaStatusMsg'></span></div>
<div class="g-recaptcha" data-callback="captchaSolved" data-sitekey=""></div>
</form>
```

### DIFF
--- a/www/capreq
+++ b/www/capreq
@@ -1,6 +1,6 @@
 <?php
 
-// Generic xml http request CAPTCHA handler.
+// Generic JSON http request CAPTCHA handler.
 //
 // This coordinates captcha handling with other scripts via the $_SESSION
 // context.  See captcha.php for an example of use.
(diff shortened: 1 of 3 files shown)


dfabulich:
I agree that `submitCaptchaForm` is dead code. I added a `debugger` statement to it and tried using it on `showuser` and on `ifarchive-upload`, but it never triggered.

I also added captcha keys in the `main` branch and tested it on `showuser` and `newuser`, verifying it that it won't let you create a user until you solve a captcha. I also tested failure by forcing a failure in `capreq` line 58.

PR #1015, opened 2024-11-07
Add an admin tool to fix UTF8 entities

### LINKED ISSUE #1002: Convert the HTML entities in the database to utf-8
> This can be done programmatically by going over all the tables, finding the text-like columns, and scanning rows one by one.
> 
> In PHP the function is `html_entity_decode`, and in Python it's `html.unescape`. I don't think node has it built-in.
> 
> I can write it if you want.
> 
> Once that's done, only rich text elements should inject such HTML entities into the page. `htmlspecialcharx` should otherwise NOT leave `&#nnnn;` sequences intact.

_Originally posted by @salty-horse in https://github.com/iftechfoundation/ifdb/issues/998#issuecomment-2456259315_

dfabulich:
Fixes #1002 

It runs pretty slowly, especially around reviews, due to the `reviews_update` trigger, which pointlessly regenerates the game ratings materialized view on every update.

But "slowly" means under a minute on my laptop, so I didn't think it was worth it to optimize (especially since I hope to only run this once in production).

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -2147,6 +2147,139 @@ else if (isset($_REQUEST['users'])) {
 
     echo "<p><hr class=dots><p>";
 
+} else if (isset($_REQUEST['utf8entities'])) {
+    function find_columns_to_fix($db, $table, $columns) {
+        $columns_to_fix = [];
+        foreach ($columns as $column) {
+            $result = mysqli_query($db, "SELECT count(`$column`) from `$table` where `$column` like '%&#%;%'");
+            [$count] = mysqli_fetch_row($result);
+            if ($count) $columns_to_fix[] = $column;
+        }
+        error_log("$table " . json_encode($columns_to_fix));
+        return $columns_to_fix;
+    }
+
+    $result = mysqli_query($db, "SELECT table_name FROM information_schema.tables
+        WHERE TABLE_SCHEMA = 'ifdb'
+        and TABLE_TYPE='BASE TABLE'
+        and table_name not like '%_mv'
+        order by table_name
+    ");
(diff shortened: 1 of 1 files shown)


salty-horse:
About "only once":

There are some "rich text" fields like reviews. Should users be allowed to enter Unicode entities, or should fixDesc convert them?

dfabulich:
> There are some "rich text" fields like reviews. Should users be allowed to enter Unicode entities, or should fixDesc convert them?

IMO, if users enter `&#128556;` in their review, we should store it and display it as 😬, but I'm not strongly committed to that point of view.

dfabulich:
… and that's not working right now. 😖

salty-horse:
> … and that's not working right now. 😖

Are you looking at the database, or the visuals?

Don't forget this still exists:
```php
function htmlspecialcharx($str)
{
    // first do the ampersands
    $str = preg_replace("/&(?!#[0-9]{1,7};)/", "&amp;", $str);
```

salty-horse (approved):
I'd be afraid to click "Fix" without seeing a preview of everything it found, and a before/after. :)

Seems to work fine locally, and I can't think of a reason not to run it, or anything it might break.

If you want to keep it around and "maintain" it, I'd turn the primary keys from a series of conditionals into an array, but it's really not needed.

salty-horse:
> IMO, if users enter `&#128556;` in their review, we should store it and display it as `&#128556;`, but I'm not strongly committed to that point of view.

More accurately, it should be rendered in HTML as `&amp;#128556;`. And for that `htmlspecialcharx()` should this part of its regex `(?!#[0-9]{1,7};)`. But really, that code should just be replaced with `htmlspecialchars()`.

Issue #383, opened 2024-09-25
Automatically check "I've played it" when you rate/review a game


dfabulich:
Lots of people have posted ratings/reviews without checking the "I've played it" box. This leads to bugs where we undercount the number of people who have "played" a game iftechfoundation/ifdb-suggestion-tracker#339 or where searches for `played:no` returns a bunch of games you've rated/reviewed but not checked the box. iftechfoundation/ifdb-suggestion-tracker#494 

We had a thread about this https://intfiction.org/t/ifdb-played-yes-and-ratings-reviews/71269 and the general consensus was that we should automatically check the "I've played it" box when you rate/review a game.

salty-horse:
Similarly, marking "I've Played It" should unmark "It's on my wishlist".

PR #1018, opened 2024-11-07
Estimated Play Time

alice-blue:
This is a draft. It should allow users to vote for a time on a game, save it, and remove it, and see an estimated time for a game displayed on the game page. (The estimated time is the median vote, with the minutes rounded to the nearest 5 minutes if the game is over an hour).

This PR does not attempt to address searching by game length.

The following games have some sample times already added in my-schema-2:

Ninja (by Panks)
The Tempest (by Grigg)
Four Seconds (by Reigstad)

**Known problems/questions**

* <s>When you are logged out, most of the game page doesn't appear.</s> @dfabulich fixed this.

* <s>The estimated time for a game could use tidying up (e.g. singular or plural for "hours," "minutes" depending on the number.)</s> Should be fixed.

* <s>Need to figure out how to display the other "time votes" for a game. Right now they are just kind of thrown in there with details tags. Where should they be displayed? Or maybe I should just leave them out and save them for a future PR.</s> They're in a bulleted list now.

* When you submit a new time, it currently does not display the new time on the page. (It does tell you "Saved," though.) I thought about displaying the new time and putting the "Remove Time" button back on, but 1) I'm not sure how and 2) It might be misleading to have the user's time keep updating when the overall estimated time won't update until you refresh the page. Anyway, I'm not sure that the way it is right now (not displaying the player's updated time or switching back to the "remove time" controls until you refresh the page) is actually a problem, but I'm explaining why it is the way it is.

* Related to above: I'm not sure if it is possible to rely on a callback function to detect if the update was actually successful. At one point I was trying to make an update to a time, but I had messed up a field name somewhere, and I got some kind of error message, but the callback function triggered anyway. 

* The number fields might allow things they are not supposed to allow. They seem to allow plus signs. And minus signs, which I tried to eliminate with a regex I found. I think I got it to reject decimals IIRC.

* Do we need more validation on the server side?

* <s>It seems like sometimes the page doesn't update with a just-added time when I refresh it. I noticed this with one of the new games that appeared on the front page, not the ones that already have sample times entered. Is this a caching issue? Is there anything we can/should do about it?</s> @dfabulich fixed this

* Since JavaScript is used to hide two of the three sections of the Game Length panel at any given time, should they all be set to "display:none" by default in case someone doesn't have JavaScript enabled? (How? Would this need another nonce?)

* It would be helpful in admin mode to show the userid next to each time vote. Right now, it does not ever show the userid.


dfabulich:
I filed https://github.com/alice-blue/ifdb/pull/1 to resolve the conflict and https://github.com/alice-blue/ifdb/pull/2 to fix the logged out issue.

alice-blue:
Thanks, @dfabulich !

FYI, I'm adding more problems/questions to the original post as I think of them.

www/viewgame
```diff
@@ -1326,9 +1337,230 @@ function updateUnwishList(id, stat)
...
+
+	    echo "<div id='timeSavedSection'>"
+        . "<p id='timeSubmittedMessage'>Your time has been submitted.</p>"
+	    . "</div>";
+
+
+// ---------------------Javascript for the Game Length Panel---------------------
+?>
+
+<script type="text/JavaScript" nonce="<?php global $nonce; echo $nonce; ?>">
```
salty-horse, line 1399: no need for `type="text/javascript"` - it's the default. And all the code can be in a single <script> tag, no need to create several of them, at least in this case.

Also consider moving all JS for this feature into a separate file.

www/viewgame
```diff
@@ -1326,9 +1337,230 @@ function updateUnwishList(id, stat)
...
+    var new_minutes = document.getElementById('minutes').value;
+    if (new_hours == "") {    // If the hours field was blank, interpret it as 0
+        new_hours = 0;
+    }
+    if (new_minutes == "") {  //If the minutes field was blank, interpret it as 0
+        new_minutes = 0;
+    }
+    var hours_are_digits = isOnlyDigits(new_hours); 
+    var minutes_are_digits = isOnlyDigits(new_minutes);
+    if (hours_are_digits == false || minutes_are_digits == false) {
```
salty-horse, line 1415: use `!variable` instead of `variable == false`

www/viewgame
```diff
@@ -1719,7 +1951,81 @@ if ($ratingCntTot == 0 && $memberReviewCnt == 0) {
...
+       return $median;
+       }    
+   }
+
+// **** Find Median Function Ends Here
+
+//***** Query for Times sectionbegins here */
+$result = mysqli_execute_query($db, "select timevote from playertimes where gameid = ?", [$id]);
+if (!$result) throw new Exception("Error: " . mysqli_error($db));
+$alltimes= array();
```
salty-horse, line 1984: use `[]` instead of `array()` in new code

www/viewgame
```diff
@@ -1719,7 +1951,81 @@ if ($ratingCntTot == 0 && $memberReviewCnt == 0) {
...
+while ($row = mysqli_fetch_assoc($result)) {
+    $time=$row['timevote'];
+    $alltimes[]=$time;
```
salty-horse, line 1987: If `$time` is ununsed, this can be single line: `$alltimes[] = $row['timevote'];`

www/viewgame
```diff
@@ -1719,7 +1951,81 @@ if ($ratingCntTot == 0 && $memberReviewCnt == 0) {
...
+$estimated_hours = 0;
+$estimated_minutes = 0;
+$mediantime=findMedian($alltimes);
+$median_hours_and_minutes=convertToHoursAndMinutes($mediantime);
+$estimated_hours=$median_hours_and_minutes['hours'];
+$median_minutes=$median_hours_and_minutes['minutes'];
+if ($median_minutes != 0) {
+    if ($estimated_hours >= 1 ) {
+        // If the game is over an hour, round the minutes to the nearest 5 minutes. 
+	    $estimated_minutes = round( $median_minutes / 5 ) * 5;
```
salty-horse, line 2000: Bad indentation (And I think it's tabs instead of spaces...? The PHP code uses spaces. Do a search-and-replace to find more)

www/viewgame
```diff
@@ -1719,7 +1951,81 @@ if ($ratingCntTot == 0 && $memberReviewCnt == 0) {
...
+            }
+            if ($estimated_minutes >= 1) {
+                echo $estimated_minutes . " minutes ";
+             }
+            echo "(based on " . count($alltimes) . " votes)</span></summary><br>";
+            echo "[Votes: ";
+            foreach ($alltimes as $t) {
+                echo " " .$t . " minutes.";
+            }
+            echo " Median: " . $mediantime . " minutes]<br>";
```
salty-horse, line 2023: You can put the `$t` and `$mediantime` inside the strings:
```php
echo "$t minutes."
```

sql/patch-schema-2.sql
```diff
@@ -76,3 +76,55 @@ ALTER TABLE `userfilters` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unico
...
+--
+-- Table structure for table `playertimes`
+--
+
+DROP TABLE IF EXISTS `playertimes`;
+CREATE TABLE playertimes (
+  id INT AUTO_INCREMENT,
+  gameid VARCHAR(32) NOT NULL,
+  userid VARCHAR(32) NOT NULL,
+  timevote INT(5) unsigned not null,
```
salty-horse, line 90: might be good to rename `timevote` to mention the actual time unit used, to avoid conversion errors in the future. `timevote_mins`? `minsvote`?

salty-horse (commented):
1) The line "X hours Y minutes (based on 7 votes)" should have some kind of description before it, like "Average playing time".

2) There should be a way to enter your own time next to that line, because users will likely look at it and say "that's way off", and want to edit it. That's where I looked for it instead of the sidebar.

3) For the breakdown of votes, I suggest putting it in dedicated predfined buckets, or a histogram like the reviews, in chunks of 30 minutes. If using text, maybe only mention half-hours in the beginning, then move to hours, and only mentions each range if it votes.

> Less than 30 minutes: X otes
> 30 minutes–1 hour: X votes
> 1 hour–2 hours: X votes
> 2 hours–3 hours: X votes
> 4 hours–5 hours: X votes

4) In the sidebar, maybe not show the time input unless you've marked "played it"?

salty-horse:
Please add the new table to `delgame`, which deletes games, and `close_user_acct()` in `util.php`, which deletes users.

sql/patch-schema-2.sql
```diff
@@ -76,3 +76,55 @@ ALTER TABLE `userfilters` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unico
...
 ALTER TABLE `userScores_mv` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 ALTER TABLE `wishlists` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+
+
+--
+-- Table structure for table `playertimes`
+--
+
+DROP TABLE IF EXISTS `playertimes`;
+CREATE TABLE playertimes (
```
salty-horse, line 86: add index with two columns: gameid and userid

www/settime
```diff
@@ -0,0 +1,132 @@
...
+// make sure the game is valid
+$qgame = mysql_real_escape_string($game, $db);
+$result = mysql_query("select id from games where id = '$qgame'", $db);
+if (mysql_num_rows($result) == 0)
+    sendResponse("Not Saved", "The specified game was not found.", false);
+
+// if there's an existing time for this user for this game, simply update
+// the time; otherwise add a new row
+
+$result = mysql_query("lock tables playertimes write", $db);
```
salty-horse, line 70: why the lock?

www/settime
```diff
@@ -0,0 +1,132 @@
...
+if ($result) {
+    $result = mysql_query(
+        "select id, timevote from playertimes
+         where gameid='$qgame' and userid='$userid'", $db);
+}
+if ($result && mysql_num_rows($result)) {
```
salty-horse, line 76: Maybe there's no need to check the current vote.

I think you can do the delete/update/insert based on the gameid and userid, instead of the row ID.

You can try the delete, and check the result to realize if it existed or not, and base the error/success message on that.

And insert/update can be done in a single "upsert", and use the result to find out which it was, to build the correct success message. (I don't know how to do that in PHP, but I assume there's a way)

www/viewgame
alice-blue: Yes, I'm planning to get rid of the extra script tags once everything is more or less in place. I was running into problems where I would move a chunk of code and it would break things.

www/settime
alice-blue: That's copied from the setrating page. I don't know much about it.

salty-horse:
Time-to-play votes should appear in these additional pages:
* Members who've played game X (`/whoselist?played&game=X)`: https://ifdb.org/whoselist?game=aearuuxv83plclpl&played
* Your Played Games (`/playlist?type=played`): https://ifdb.org/playlist?type=played

www/settime
salty-horse: I don't know why it's there either :D @dfabulich ?

www/settime
alice-blue: Almost everything on this page is copied from setrating, fyi. :)  I do not claim to understand it all.

(Maybe some things that were needed for setrating won't apply here? Setrating deals with both reviews and ratings, but here we are only dealing with times.)

salty-horse: I'll submit a PR later converting it to the new JSON approach, and trying out my suggestion.

salty-horse:
Should this be a field in the playedgames table instead of its own table?

alice-blue:
> Should this be a column in the playedgames table instead of its own table?

@dfabulich what do you think? I don't think I even knew there was a playedgames table. I'm not sure what's in it.

salty-horse:
When you mark "I played this" on a game, it's stored there.

Since I think it makes sense to only let you suggest times for games you've played, that same record can be used to store extra data.

www/settime
dfabulich: That lock has been in `setrating` since we inherited the code from MJR. Reading the comments, it looks like he was using the lock to do an "upsert", which nowadays we would do with `INSERT … ON DUPLICATE KEY UPDATE`.

dfabulich:
The thread about this https://intfiction.org/t/time-to-completion-where-to-put-it/72078 briefly discussed putting it on the `reviews` table, or giving it its own table. Putting it in the `playedgames` table is another option, I suppose.

One weirdness about that is that we never got around to automatically checking "I've played it" when you rate/review a game. #383

Another weirdness: if you _uncheck_ "I've played it" after you've submitted a time vote, it deletes the row from `playedgames`. Would it delete your time vote, too? We don't treat ratings or reviews like that. We would never delete a written review just because you unchecked the "I've played it" box.

For reasons like this, I think it makes sense that @alice-blue opted for a separate table, because it ensures we can do whatever we want to do in the UI.

dfabulich:
More broadly, a lot of @salty-horse's comments here would be solid improvements, but it's not clear to me which of these we feel are needed before publishing, and which could be done in separate follow-up cleanup PRs.

(I'm sensitive to @alice-blue's remark on the forum that they might be running out of steam.)

Once we show it to the public, there's going to be _even more_ feedback, and even more little things to fix?

What do we think _must_ be done before we go live on ifdb.org, gathering time votes and gathering public feedback on the design? (The worst case would be spending a bunch of time "cleaning up" code that we then delete or completely reimplement based on forum feedback.)

At a minimum, I think we do have to confirm that we've picked the "right place" to store the data, i.e. in `playertimes` or in another table.

And ideally we'd not deposit a bunch of technical debt on ourselves, where we'd ship this thing and then file dozens of bugs (especially [code cleanup](https://github.com/iftechfoundation/ifdb/issues?q=is%3Aissue+is%3Aopen+label%3A%22code+cleanup%22) bugs) to fix "someday" that we never get around to.

salty-horse:
I'm OK with the separate table. A safe approach.

A perfect UI isn't that important. But for the initial release, there should be a news item, and maybe some call-out on game pages to submit this info when you mark a played game, or visit a page that you already marked as played, with a highlight on the new field, so it suggests interacting with it.

This requires moderate user input to become useful.

sql/patch-schema-2.sql
alice-blue: Maybe time_in_minutes? Vote_in_minutes? I did wonder whether we might eventually add any other columns to the table, and what we might need to distinguish it from (e.g. if we store the median of the game as well, in which case "time_in_minutes" might be too generic, but I guess we could change that later. Maybe "vote_in_minutes" is safer.

dfabulich:
We'll definitely do site news for it.

In the current implementation, the time vote stuff is showing regardless of whether you've clicked "I played it." I think that's how it's gotta be unless/until we do #383, because otherwise I'm afraid lots of people won't discover how to submit time votes, because they'll simply rate/review without realizing that they also need to check the box to submit a time vote.

So, here's what I think:

@alice-blue wrote on the forum:
> I think we/I should at least fix the typos, the singulars/plurals, figure out where and how to display the votes, and decide which table we want to store the values in. Also I’m thinking of changing the field name. (Beyond that, I’ll leave it up to you all to decide what’s the minimum necessary.)

I think we're 👍  for storing this in the `playertimes` table.

re: "figuring out where and how to display the votes," I think we're going to have to deploy to the public to get a clearer sense of where to display the votes. In the meantime, showing them in the top section in a `<details>` is "good enough to ship," IMO, with the understanding that we'll likely change it, but it's impossible to say for sure _how_ we'll change it.

(But, if you feel like you want to gather more feedback from the advisors group, feel free!)

dfabulich:
In practice, I think once we deploy this, we should just file a bunch of bugs about time voting, and pick 'em off one by one.

And, uh, here's another enhancement request: Show the time vote form on the `/review` page, so if a user navigates there, they can do it all in one place: rate, review, submit a time vote. (Again, this is _not_ required to ship, but I think it would be real nice.)

sql/patch-schema-2.sql
alice-blue: I changed this to vote_in_minutes.

www/util.php
```diff
@@ -2895,6 +2895,44 @@ function collapsedAuthors($authors) {
...
+    $m=$hours_and_minutes['minutes'];
+    if ($h >= 1) {
+        $text = $h . " ";
+        if ($h == 1) {
+            $text .= "hour";
+        } else {
+            $text .= "hours";
+        }
+        if ($m >= 1) {
+            $text .= " and";
```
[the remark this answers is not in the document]
alice-blue: Should be fixed.

www/viewgame
```diff
@@ -1326,9 +1337,230 @@ function updateUnwishList(id, stat)
...
+   
+
+
+//-----------------------Game Length Panel Begins----------------------------
+        echo "<div class=indented>";
+        echo "<div id='gameLengthPanel' class='viewgame__playedSection'>"
+        . "<span id=\"gameLengthCaption\"><b>Game length:</b></span>";
+
+        echo "<div id='submitTimeSection'>"
+        . "<span id='howLong'>Vote on long it takes to finish this game.</span><br>"
```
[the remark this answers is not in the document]
alice-blue: Should be fixed.

alice-blue:
I think I fixed the things that are actually visible to users. One thing I wasn't sure about, @salty-horse , is that you said you weren't a fan of "Game Length" on the sidebar, but I wasn't sure if you didn't like the wording, or you didn't think it should be there at all. For the moment I changed it to say "How long is this game?" instead.

alice-blue:
Oh, but I didn't put in an error if you try to add a time without adding it to the play list. I'll take a look and see if that looks like a quick fix or not.

alice-blue:
Should the listed votes be sorted by length of time?

salty-horse:
I only meant the wording of "Game length" - it's suggest the game has a prescribed length, when actual playing times will vary from person to person.

The list of votes will become unusable if there are a lot of votes with different times, even if they're off by minutes. I think it's important to put them into "buckets" of 30 minute chunks, and sort by number of votes. You can decide if it's important for the first release or not.

alice-blue:
Does "How long is this game?" work for the heading, or would you prefer something else?

salty-horse:
My suggestions are "Average playing time" or "Estimated playing time", but whatever is fine. This is nitpicking and can be changed later based on feedback.

Sorting should be by descending votes, I think, and then by ascending time. It can be confusing, but the most voted stuff should be the most prominent.

dfabulich:
I hypothesize that the long-term answer will be to show time votes in a histogram, but it probably won't be useful to do a histogram until there are more than a dozen votes on a game, so there's no sense in fussing the details on displaying/sorting votes right now.

dfabulich:
FYI, I'm not keeping track of this PR commit-by-commit; @alice-blue I invite you to click the ♺ "Re-request review" button next to my name in the "Reviewers" section at the top of the page when you feel like you've made the changes you want to make.

alice-blue:
Ok. Assuming I didn't break anything, I think the minimum things are finished. Anything I do from here on out would be more minor.

alice-blue:
Oops. I did break something. All three sections of the time controls are now visible at the same time.

alice-blue:
I think I fixed the error.

salty-horse:
I'll convert `settime` to JSON tomorrow.

alice-blue:
I tried to work out how to add display:none to the css for those three divs, but I'm not sure I got it right, or how I would even tell (would I need to disable JS to test it)? Anyway, that's not in this PR.

dfabulich:
I filed a couple more PRs on this: https://github.com/alice-blue/ifdb/pull/3 and https://github.com/alice-blue/ifdb/pull/4

alice-blue:
@dfabulich I went ahead and merged one of the commits without testing it first (maybe I shouldn't have) and now I'm trying to test the other one and I'm getting a database error message:

An error occurred accessing the user database. Please try again. If the problem persists, please [contact us](https://vigilant-sniffle-7v7444wjxqqh7gq-8080.app.github.dev/contact) to let us know about the problem.

I tried docker compose down, up, etc.

dfabulich:
You have to `./prepare_dev_environment.sh` every time you change `patch-schema-2.sql`, _then_ `docker compose down` and `docker compose up`.

… but it turns out, I didn't test it either, because, when I tried that, I get this error on the Docker console:

```
db-1   | ERROR 1062 (23000) at line 118: Duplicate entry '0000000000000001-59g5czw7izz7aoip' for key 'user_game'
```

That is to say, good job index, because it prevented the DB from accidentally submitting two time votes for the same game for the same users. I've fixed it in https://github.com/alice-blue/ifdb/pull/5.

alice-blue:
@dfabulich I just added another commit to deal with singular/plural for the estimated median time. (It doesn't use the same function as the other times, because of the rouding-to-nearest-5-minutes thing.)

dfabulich:
I think this is in pretty good shape.

> I'll convert settime to JSON tomorrow.

@salty-horse Let's merge this after you file a PR to port `settime` to JSON. (Or should we merge it now, and you'll file a PR after we merge?)

alice-blue:
@dfabulich, if I want to make any other changes that were brought up (like fixing indentation, CSS, etc.), should I keep committing to this branch, or just leave this one alone and start a new one after it's merged?

dfabulich:
oops, I accidentally force pushed your branch. I'll fix it

dfabulich:
I undid my mistake. (I didn't realize I had write access to the `alice-blue/ifdb` repo… 🤔)

In the meantime, here's how I propose that we continue.

1. I've pushed a copy of the `game_length` branch to this repo https://github.com/iftechfoundation/ifdb/tree/game_length I can merge that at any time.
2. I recommend that you start doing a number of smaller PRs targeting the https://github.com/iftechfoundation/ifdb/tree/game_length branch. (When you create the PR, you can select the target destination branch.) Smaller PRs are easier to review and merge.
3. You can also feel free to start filing issues regarding estimated play time, so that you or anybody can work on them.
4. @salty-horse can do a PR targeting https://github.com/iftechfoundation/ifdb/tree/game_length to convert to JSON
5. When that's done, I'll merge our copy of `game_length`, in whatever state it's in. If you have some open PRs targeting the `game_length` repository, we'll make them target the `main` branch after `game_length` merges.

We'll try to make sure our copy of the `game_length` branch is basically shippable until it merges.

So, if you're feeling out of steam, feel free to take a break. If you're energized, you can do work targeting our copy of the `game_length` branch.

sql/patch-schema-2.sql
```diff
@@ -76,3 +76,58 @@ ALTER TABLE `userfilters` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unico
...
+  id INT AUTO_INCREMENT,
+  gameid VARCHAR(32) NOT NULL,
+  userid VARCHAR(32) NOT NULL,
+  time_in_minutes INT(5) unsigned not null,
+  PRIMARY KEY (id),
+  KEY `gameid` (`gameid`),
+  KEY `userid` (`userid`),
+  UNIQUE KEY `user_game` (`userid`, `gameid`)
```
salty-horse, line 94: I think it makes more sense for the unique key to have the game first, to match the order of the columns.

And then there's no need for the KEY on `gameid`, since any queries on just `gameid` can reuse the index `gameid, userid`.

dfabulich (approved):

PR #1024, opened 2024-11-07
Query IFWiki from the browser, remove ifwiki-check

### LINKED ISSUE #606: Avoid querying IFWiki whenever a game page is loaded

salty-horse:
Also changed "IF Wiki" on the page to "IFWiki", which is the official spelling.

This probably fixes #606 - it's still a query, but the IFDB server is oblivious :D


dfabulich:
IMO this doesn't fix #606, because we're still wasting the user's bandwidth (and now we're using a _different_ IFTF server's resources). It's still good cleanup code, so I'll merge it.

PR #1026, opened 2024-11-07
Remove special cases for &#nnnn; entities

dfabulich:
Not actually ripping out `htmlspecialcharx` right now, because that would cause a ton of conflicts.

### DIFF
--- a/www/util.php
+++ b/www/util.php
@@ -155,19 +155,11 @@ function approx_utf8($str)
 
 // ------------------------------------------------------------------------
 //
-// htmlspecialchars() replacement.  This converts everything that
-// the standard htmlspecialchars() does, *except* that it leaves
-// &#nnnn; sequences intact.  Since we don't store Unicode directly,
-// this provides a rudimentary way of storing these characters.
+// htmlspecialchars() replacement. TODO delete
 //
 function htmlspecialcharx($str)
 {
-    // first do the ampersands
-    $str = preg_replace("/&(?!#[0-9]{1,7};)/", "&amp;", $str);
-
-    // now do the rest of the characters and return the result
-    return str_replace(
-        array('"', '<', '>'), array('&quot;', '&lt;', '&gt;'), $str);
+    return htmlspecialchars($str);
 }
 
 // sometimes we want to display long file names containing underscores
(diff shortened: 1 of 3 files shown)


www/util.php
```diff
@@ -1398,8 +1386,7 @@ function summarizeHtml($str, $maxlen)
                 if (strncasecmp($entTxt, "lt;", 3) == 0
```
salty-horse, line 1386: a line above here is `$entTxt = substr($str, $ofs + 1, 7);`
I think the 7 can be reduced to 5, which is the farthest ahead this little test cares about.

It was 7 in the original code import, when the regex was `{1,5}`. When I changed the regex to `{1,7}`, I should have increased that number to 9. It's nice to find bugs after they're obsolete.

salty-horse (changes requested):
More places to edit:

searchutil has an `$exacts = false;` and `if ($exacts)` block.
```
www/searchutil.php:309:    // list of exact phrases for &#xxxx; sequences - none so far
www/searchutil.php:451:        // add the exact matches for &#xxxx; phrases
```

And util.php has an outdated comment mentioning &#xxxxx, and substr that should change to a 5 instead of 8:
```
www/util.php:1157:            // if it's &lt;, &gt;, &#xxxxx;, &quot; or &amp;, leave it;
```
```php
case '&':
    // if it's &lt;, &gt;, &#xxxxx;, &quot; or &amp;, leave it;
    // otherwise convert the & to &amp;
    $therest = substr($desc, $ofs + 1, 8);
    if (strncasecmp($therest, "lt;", 3) == 0
        || strncasecmp($therest, "gt;", 3) == 0) {
        $ofs += 3;
    } else if (strncasecmp($desc, "amp;", 4) == 0) {
        $ofs += 4;
    } else if (strncasecmp($desc, "quot;", 5) == 0) {
        $ofs += 5;
    } else {
        // not recognized - make it an explicit &amp;
        $desc = substr_replace($desc, "&amp;", $ofs, 1);
        $ofs += 4;
    }
    break;
```

salty-horse:
You can also delete this from `www/search`, unless you want it in a separate commit. The second line *is* a special case for these entities:
```php
// Convert database row from latin1 to UTF-8
$g = rss_encode($g);
$g = html_entity_decode($g);
```

salty-horse (changes requested):
Sorry, missed this 7 that should be 5:
```
else if ($c == '&')
{
    // if it's &lt;, &gt;, or &amp;, it's an entity
    $entTxt = substr($str, $ofs + 1, 7);
    if (strncasecmp($entTxt, "lt;", 3) == 0
        || strncasecmp($entTxt, "gt;", 3) == 0
        || strncasecmp($entTxt, "amp;", 4) == 0
        || strncasecmp($entTxt, "quot;", 5) == 0) {

```

salty-horse (approved):

Issue #1035, opened 2024-11-08
A new time vote doesn't always show up when refreshing the page


alice-blue:
When I add a new time vote and refresh the page, it seems like sometimes the page doesn't update to show the correct number of votes. Is this a caching issue? Is there anything we should do about it?

dfabulich:
I believe the issue is that the time votes section only appears when there's at least 1 rating or at least 1 written review.

In the local dev DB, the newest game is "REPEAT IT BACK TO ME" and it has no ratings/reviews.

* If you submit a time vote for it, you won't see it above the descirption; it just says, "No reviews yet."
* If you submit a rating, it says:
    > ★★★★★ (based on 1 rating)
    > Estimated play time: 3 hours (based on 1 vote)
    > No reviews yet
* If you remove the rating, the "Estimated play time" disappears.
* If you write a review with no star rating, it says:
    > Estimated play time: 3 hours (based on 1 vote)
    > 1 review
* If you delete the review, the "Esttimated play time" disappears again.

alice-blue:
I think the whole section to display the time is (unintentionally) inside of an "else" clause in the rating/review section.

dfabulich:
I'm fixing it

PR #1039, opened 2024-11-08
Show median when there are 0 ratings and 0 reviews

dfabulich:
Fixes https://github.com/iftechfoundation/ifdb/issues/1035

@alice-blue

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1922,7 +1922,7 @@ if (!isEmpty($genre)) {
     if ($detailView) {
 ?>
          <p>
-         <span class=details>
+         <div class=details>
             <?php
 
 $embargoCnt = 0;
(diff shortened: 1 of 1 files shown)


alice-blue:
Thanks for fixing this. Once this is merged, I want to do a PR to rewrite some of my other code in this section.

www/viewgame
```diff
@@ -2004,48 +1982,73 @@ $median_minutes=$median_hours_and_minutes['minutes'];
...
-	    $estimated_minutes = round( $median_minutes / 5 ) * 5;  
+        $estimated_minutes = round( $median_minutes / 5 ) * 5;
     } else {
         // If the game is under an hour, round the minutes to the nearest whole number.
         $estimated_minutes = round($median_minutes);  
     }
 }
 
-
+function displayMedian($estimated_hours, $estimated_minutes, $alltimes) {
```
salty-horse, line 1992: Mention "Time" in the function name

salty-horse (approved):

www/viewgame
```diff
@@ -2004,48 +1982,73 @@ $median_minutes=$median_hours_and_minutes['minutes'];
...
+    echo "<img src=\"/img/blank.gif\" class=\"star0\" title=\"No Ratings\">
+       No reviews yet";
+    if ($embargoCnt == 0)
+        echo " - <a href=\"review?id=$id&userid=$curuser\">be the first</a>";
+} else if ($ratingAvg == 0 && $memberReviewCnt == 0) {
+    echo "<a href=\"viewgame?id=$id&ratings\" "
+        . "title=\"View all ratings and reviews for this game\">"
+        . "$ratingCntTot rating" . ($ratingCntTot != 1 ? "s" : "")
+        . "</a><br>";
+    displayMedian($estimated_hours, $estimated_minutes, $alltimes);
```
dfabulich, line 2038: This is a weird passage that probably can't happen in real life. I might delete it later in a separate PR.

The difference between `$ratingCntTot` and `$ratingCntAvg` is that you can publish a written review and check a box, "Omit my rating from the game's average." In that case, the rating is part of the "total" but it's not part of the "average," not part of the score at the top of the page.

It's possible for all of the ratings to be omitted, in which case, `$ratingAvg` will be null (which is `== 0`).

The first branch of this `if`/`else` chain considers the case where there are 0 ratings and 0 reviews. The third considers where there's at least one rating or at least one review, and handles the case where the `$ratingAvg` is null.

This middle section imagines a case where `$ratingAvg` is null, but `$memberReviewCnt` is 0. That's not something that can happen just by clicking around in the web UI, but it is representable in the database, if you write a review with an omitted rating, and then update the DB directly to null out the `review`.

Issue #646, opened 2021-03-21
Use HTML Doctype


curiousdannii:
It's very important to add the HTML Doctype so that browsers will display the site in standards mode. Any CSS changes we make before then are at risk of being wrong.

salty-horse:
Can we try this on the dev website?

salty-horse:
I tried it, and the game page looks different enough to require a bit of work (a few elements move around, some are shorter than intended) , but I'm not sure how much work. The home page looks pretty much the same except the shorter header.

Probably worth addressing sooner rather than later.

`www/pagetpl.php` is the place to change:
```patch
+<!doctype html>
 <html lang="en">
```

salty-horse:
Another issue that affects many places is tables not respecting the `max-width` set on the "main area" of the page.

The issue is that `max-width` like so:
```
p, ul, ol {
    max-width: 60ch;
}
```
and those tables are inside `<p>`. In standards mode, `<p>` can't contain `<table>`, so it it's placed after it, out of the effect of `max-width`.

This can be seen in:
* The [File Formats](https://ifdb.org/fileformat) page, where the text extends far beyond `max-width` set by the `<p>` it's  in. Also, in standards mode, tables aren't allowed in `<p>`, so 
* The game page, where the covert art is further to the right than the rest of the text.

I think the solution is going over all the pages with tables, and setting that max-width on the larger containers. For example, in `viewgame`, that would be `viewgame__mainSummary`.

BTW, the CSS 2.1 standard [says](https://www.w3.org/TR/CSS21/visudet.html#propdef-max-width):
> In CSS 2.1, the effect of 'min-width' and 'max-width' on tables, inline tables, table cells, table columns, and column groups is undefined.
Is this still relevant in modern browsers?

Is there an easy fix for this other than going over all the tables in the codebase?

salty-horse:
Can't just set `table { max-width: 60ch; }` because pages like http://localhost:8080/personal use a table to layout both the main area and the sidebar.

salty-horse:
Hit another difference:

In quirks mode, images are `display: block` by default, and `display: inline` in standards mode. The inspectors in both Firefox and Chrome will always report it as `display: inline`, whether you're in quirks mode or not.

salty-horse:
Popups, like the external link editor on the on the "Edit game" page, have a class `edit-popup-win` that sets `font-size: 85%;`.

[...]

PR #1040, opened 2024-11-08
Use semantic markup for external links

salty-horse:
As part of the preparation for #646, I'm replacing table-as-layout parts of the code.

This replaces the tables in the External Links box in the game page.

It's not pixel-perfect with the original, but I think it's fine.

I tested on:
* [Game page `viewgame`](https://ifdb.org/viewgame?id=jephaktzt5svmav1&downloads)
* [Game page with `&downloads`](https://ifdb.org/viewgame?id=jephaktzt5svmav1&downloads) - here it's more noticable there's no indentation of a few pixels on the left margin, but I don't think it's needed.
* [The CSS preview page `cssthumb`](https://ifdb.org/cssthumb)

It might break some custom stylesheets that depend on tag names such as "table". Is that important?


dfabulich:
> It might break some custom stylesheets that depend on tag names such as "table". Is that important?

Not to me! IMO, the only important custom stylesheet is Midnight (because that's the only way for users to force a preference to opt-in to dark mode), but it's one line now, nothing but `@import url("/ifdb.css");`, with "Force Dark Mode" set to Dark.

salty-horse:
This broke links without icons. I'll fix it.
link

salty-horse:
Fixed in eb0bfbe3910c073b678db01f8aece8a27c5d4fc8

PR #1046, opened 2024-11-08
Remove unneeded indexes for playertimes table

dfabulich:


salty-horse (approved):
I guess it can be useful to show times in the "Your Played Games" page, but then the query is by `gameid IN ()` so no index is need for just the user.

PR #1054, opened 2024-11-08
Use the convert time to text function for displaying the estimated median time

alice-blue:
I tried to separate the parts of the convert time to text function so that it could also be used in displaying the estimated median time.

However, this might have introduced a bug where the reviews line ends up having too much indentation.

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1972,23 +1972,30 @@ while ($row = mysqli_fetch_assoc($result)) {
     $alltimes[]=$time;
 }
 
-
-// Find the median play time for this game, and round the minutes if necessary
-$estimated_hours = 0;
-$estimated_minutes = 0;
+// Find the median play time for this game, and separate the hous and minutes
 $mediantime=findMedian($alltimes);
 $median_hours_and_minutes=convertToHoursAndMinutes($mediantime);
-$estimated_hours=$median_hours_and_minutes['hours'];
+$median_hours=$median_hours_and_minutes['hours'];
 $median_minutes=$median_hours_and_minutes['minutes'];
+
+// For the official estimated time for a game, we we want to round to the 
+// nearest 5 minutes when the game is longer than 1 hour.
+$official_estimated_time_in_text="";
 if ($median_minutes != 0) {
-    if ($estimated_hours >= 1 ) {
+    if ($median_hours >= 1 ) {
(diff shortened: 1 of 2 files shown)


alice-blue:
I don't know why it's showing so many commits.....

dfabulich:
It was the merge base; your PR was set to target the `main` branch, which would bring in the whole thing, instead of the `game_length` branch. I just fixed it.

alice-blue:
Thank you!

www/viewgame
```diff
@@ -1971,52 +1972,43 @@ while ($row = mysqli_fetch_assoc($result)) {
...
+	$minutes_rounded_to_five = round( $median_minutes / 5 ) * 5;
+        $official_estimated_time_text = convertTimeToText(['hours' => $median_hours, 'minutes' => $minutes_rounded_to_five]);
     } else {
         // If the game is under an hour, round the minutes to the nearest whole number.
-        $estimated_minutes = round($median_minutes);  
+        $whole_minutes = round($median_minutes);
+        $official_estimated_time_text = convertTimeToText(['hours' => $median_hours, 'minutes' => $whole_minutes]);
     }
+} else {
+    $official_estimated_time_text=convertTimeToText($median_hours_and_minutes);
```
dfabulich, line 1995: When I test this, it doesn't work at all, showing no median at all in the summary.

image

This is because of variable scope. https://www.php.net/manual/en/language.variables.scope.php

Basically, the `function displayMedianTime` can't "see" any variables that aren't passed into it.

This is a common issue, so it's usually a good idea to peek at the Docker console whenever you load the page, and inspect for warnings. When I run this branch, I see this warning:

```
Undefined variable: official_estimated_time_text in /var/www/html/viewgame on line 2003
```

There are two ways to fix this:

1. We can pass `$official_estimated_time_text` as a fourth parameter to the `displayMedianTime`
2. We can move the code that computes the official text into `displayMedianTime`.

dfabulich (changes requested):

alice-blue:
@dfabulich ok, I changed it to pass  $official_estimated_time_text as a parameter to the displayMedianTime

alice-blue:
Oh. Is there a blank image there? (Are there supposed to be empty star outlines?)

```
echo "<img src=\"/img/blank.gif\" class=\"star0\" title=\"No Ratings\">
       No reviews yet";
```

PR #1061, opened 2024-11-08
Prevent mobile browser zoom

### LINKED ISSUE #1057: Prevent mobile browser zoom on login form
On iOS, when you go to the login page, we autofocus the "Email address" field, and the browser zooms in slightly. That's annoying.

https://stackoverflow.com/a/6394497/54829

> The browser will zoom if the font-size is less than 16px and the default font-size for form elements is 11px (at least in Chrome and Safari).

dfabulich:
Fixes #1057

Just getting started, because to actually fix most forms, we have to redesign them so the field label appears above the field, instead of in a table cell to the left.

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -1250,18 +1250,22 @@ span.headlineRss {
     display: none;
 }
 
-.menu-toggle-button{
+.mobile-menu #mobile-menu-toggle-button {
     display: none;
+}
+.menu-toggle-button{
+    display: flex;
     align-items: center;
     align-self: start;
     padding: 0.25em 0.75em 0.25em 0.75em;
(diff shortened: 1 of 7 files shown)


salty-horse (approved):

dfabulich:
Now login and review forms work. A few other forms luckily started working just from setting all inputs to 16px.

salty-horse:
The two links above the review box used to be aligned with its right edge, and now they moved to the right.
review

dfabulich:
Hrm. This seems to vary by browser.

dfabulich:
Fixed the review page, tested in Firefox, Chrome, and Safari.

PR #1071, opened 2024-11-08
Improve error handling for invalid taggame requests

dfabulich:
This stumped me for a while, because I was foolishly trying to submit `application/x-www-form-urlencoded` data, like the old XML API expected. Returning a real error message would have saved me time and trouble.

### DIFF
--- a/www/taggame
+++ b/www/taggame
@@ -62,16 +62,12 @@ if ($db == false)
 $xml = isset($_REQUEST['xml']);
 if (!$xml) {
     if ($_SERVER['REQUEST_METHOD'] != 'POST') {
-        http_response_code(405);
-        echo "405 Method Not Allowed";
-        exit();
+        sendResponse("405 Method Not Allowed", "Not Saved", "This API requires you to use HTTP method POST");
     }
 
     $input_json = json_decode(file_get_contents('php://input'), true);
     if (is_null($input_json)) {
-        http_response_code(400);
-        echo "400 Bad Request";
-        exit();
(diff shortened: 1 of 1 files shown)


www/taggame
```diff
@@ -62,16 +62,12 @@ if ($db == false)
 $xml = isset($_REQUEST['xml']);
 if (!$xml) {
     if ($_SERVER['REQUEST_METHOD'] != 'POST') {
-        http_response_code(405);
-        echo "405 Method Not Allowed";
-        exit();
+        sendResponse("405 Method Not Allowed", "Not Saved", "This API requires you to use HTTP method POST");
```
salty-horse, line 65: http_response_code officially only accepts integers. We should not be passing strings (rest of the code should also be fixed.

The "not saved" label is just for our UI and probably not relevant in this case, but can be left in.

salty-horse (approved):

www/taggame
dfabulich: I'm not sure I understand this comment. I think you might be suggesting that I should change this line to say `sendResponse(405, "Not Saved", "blah blah")`?

But `sendResponse` in `taggame` doesn't use `http_response_code`. It uses `header("HTTP/1.1 $status");`, which is where we _do_ have to include the status text ourselves.

I guess we could have a separate PR to replace instances of `header("HTTP/1.1 $status");` but it's probably not worth the effort.

salty-horse: You're right, I had the two approaches mixed up!

Issue #1073, opened 2024-11-08
gametags API should reject passwords submitted as GET parameters


dfabulich:
Apropos #1072, the gametags API accepts parameters either via POST or GET.

That seems fine when you're viewing public data like this:

<https://ifdb.org/gametags?id=aearuuxv83plclpl&json>

But it seems bad if you're viewing private `mine_only` data like this:

<http://localhost:8080/gametags?id=aearuuxv83plclpl&json&username=ifdbadmin@ifdb.org&password=secret&mine_only=1>

Because the request will be logged, including your password.

To try to protect users, I think we should reject GET requests with a password.

dfabulich:
@salty-horse WDYT?

dfabulich:
We deprecated gametags instead.

Issue #1072, opened 2024-11-08
taggame and gametags API accept parameters in different formats


dfabulich:
taggame and gametags now accept JSON, but in wildly different ways.

* `gametags` accept parameters either POSTed as `application/x-www-form-urlencoded` data, or in GET parameters (which will be logged 😬)
* `taggame` only accepts parameters POSTed as `application/json`

I think it's confusing to do both.

I think it's natural for `taggame` to use JSON documents (so that we can POST an array, instead of `t0` `t1` `t2` parameters as the old XML API required), so `gametags` should probably do the same thing.

dfabulich:
@salty-horse WDYT?

dfabulich:
I just filed #1081 that we should maybe have a `mytags` API.

It seems to me that `gametags` is kinda doing too much. On the one hand, you can call it with GET parameters <http://localhost:8080/gametags?id=aearuuxv83plclpl&json> and get a convenient source of public data.

But if you want to do a `mine_only` query, you really ought _not_ pass in GET parameters; you should use POST or [QUERY](https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html) parameters only, so your password won't get logged.

If `viewgame` returns public tags #1078, then maybe we don't need `gametags` at all: you could get the public tags via `viewgame` (GET, naturally) and you can get your personal tags via `mytags` (password required, so POST or QUERY only). 

And then perhaps no one would  ever want to use `gametags`, like, ever again!

salty-horse:
I agree that the current API is bad, and that passwords shouldn't be part of the URL.

Headers are also an option.

dfabulich:
We deprecated `gametags` instead.

PR #1077, opened 2024-11-08
Add JSON API documentation for gametags and taggame

dfabulich:
This is a draft while we think about what to do with #1072 and #1073.


dfabulich:
Closed in favor of #1103 and #1105

PR #1079, opened 2024-11-08
Add tags to viewgame API

### LINKED ISSUE #1078: Return tags in `viewgame` API
It's a _bit_ weird to return tags here, because tags aren't versioned with the rest of the document (see #915), but I think most users don't want to know or care about our internal schema around tag ownership. We should just put the tags in the `ifdb` section.

dfabulich:
Fixes #1078

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -699,6 +699,28 @@ if (isset($_REQUEST['ifiction']) || $json)
         $ifdb_section['ratingCountTot'] = $ratingCntTot;
     }
 
+    $result = mysqli_execute_query($db,
+        "select
+           tag,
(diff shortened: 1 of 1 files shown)


salty-horse:
This code is already in `tags.php`, in the second half of the function. It should be split into its own function and reused here.

dfabulich:
Ehhhh It's _pretty_ close, but `tags.php` has `where tag in ($questionMarks)` that makes it weird to factor out.

Obviously there's now copying and pasting between `viewgame` and `gametags`, but I'm hoping that this is a first step to removing `gametags` in favor of `mytags` #1081.

salty-horse (approved):

PR #1084, opened 2024-11-09
Improve viewgame layout in standards mode

salty-horse:
Fixes the cover art issue mentioned here https://github.com/iftechfoundation/ifdb/issues/646#issuecomment-2466082560, setting the game page header with an explicit width, instead of placing it inside a `<p>`, which is not supported.

I checked both the main page view, and specific modes like `&reviews`. It looks fine in both.

The header is now a `<section>`. Semantically, I think the review counts and playtime rows should be part of it, but the current code structure makes it difficult/less readable.

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -84,6 +84,11 @@ body {
     margin: 0px 0px 1.5em 0px;
 }
 
+:root {
+    /* The width of the main sections of a page, without any sidebars. */
+    --paragraph-width: 60ch;
+}
+
 /*
 The style for <form> areas - this is the style used for areas of the
 page where there are data entry fields, buttons, and so on.  We use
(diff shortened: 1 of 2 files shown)


dfabulich:
I'm having a hard time reproducing the bug that you're fixing.

In https://github.com/iftechfoundation/ifdb/issues/646#issuecomment-2466082560 you wrote:

> This can be seen in:
> 
> * The [File Formats](https://ifdb.org/fileformat) page, where the text extends far beyond `max-width` set by the `<p>` it's  in. Also, in standards mode, tables aren't allowed in `<p>`, so
> * The game page, where the covert art is further to the right than the rest of the text.

The file format page looks fine to me. And I see no visible effect on the games page. (But maybe the point is that this only fixes a bug that _would_ crop up in `<!doctype html>` mode…?)

image

dfabulich (commented):
Requested clarification

salty-horse:
Yes, this is fixing problems that appear when turning on Standards mode. I'm fixing them in advance, so it can be turned on for all pages at once.

Have two tabs open side-by-side, turn it on in one of them, and compare.

salty-horse:
This broke the style of the tag editor. Fixed in 79a0523d6213836216d7787bfc63de56dd2c91c8.

salty-horse:
This also broke spacing between the "External Links" and "Have you played this game?" sections. Checking...

PR #1094, opened 2024-11-09
Add a mytags API

### LINKED ISSUE #1081: Add a `mytags` API
Apropos #1072

Instead of a public `gametags` API, maybe we should rely entirely on people fetching public tags in the `viewgame` API #1078.

In order to support `mine_only` requests, perhaps we'd create a new `mytags` as an API, that only accepts `POST` parameters `username`, `password` and `id`. `mytags` would be hard coded to return `mine_only` results.

Perhaps `mytags` might graciously accept _either_ an `application/json` document _or_ `application/x-www-form-urlencoded`, but never a `GET` request, because it's supposed to manage passwords.

(Maybe it would accept a `QUERY` method, though? https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html Maybe some day, but not today…?)

dfabulich:
Fixes #1081

Try it out at http://localhost:8080/mytags?id=aearuuxv83plclpl

The browser will prompt you for basic authentication (and then keep you logged in for a while, as browsers do).

Initially, it will be blank, assuming you haven't put any tags on the game, but if you add a tag to that game, it will appear in the resulting array.

### DIFF
--- a/www/mytags
+++ b/www/mytags
@@ -1,0 +1,70 @@
+<?php
+include_once "util.php";
+
+function sendResponse($status, $errmsg, $tagInfo = null)
+{
+    header("HTTP/1.1 $status");
+    header("Expires: Mon, 26 Jul 1997 05:00:00 GMT");
+    header("Cache-Control: no-store, no-cache, must-revalidate");
+    header("Content-Type: application/json");
+
+    $responseObj = [];
+    if ($errmsg) {
+        $responseObj['error'] = $errmsg;
+    } else {
+        $responseObj['tags'] = $tagInfo;
+    }
+
+    echo json_encode($responseObj);
+
+    exit();
+}
+
+$db = dbConnect();
+if ($db == false)
+    sendResponse("500 Internal Server Error", "An error occurred connecting to the database. "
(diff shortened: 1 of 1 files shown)


www/mytags
```diff
@@ -0,0 +1,70 @@
...
+
+    exit();
+}
+
+$db = dbConnect();
+if ($db == false)
+    sendResponse("500 Internal Server Error", "An error occurred connecting to the database. "
+                 . "Please try again later.", false, false);
+
+$username=$_SERVER['PHP_AUTH_USER'] ?? null;
```
salty-horse, line 28: add space before/after `=`

salty-horse (approved):
Nice! `taggame` should also work this way

dfabulich:
I realized that we don't need a special API endpoint for this, because the search API already returns tags. Instead of `mytags?id=aearuuxv83plclpl` you can do http://localhost:8080/search?json&tag&tuid=aearuuxv83plclpl&mine:yes

I've filed #1094 to add `tuid:` and `mine:` filters to tag search, and I filed #1099 to add basic authentication support to the search API. With those in place, we no longer need this API!

PR #1101, opened 2024-11-09
Support basic authentication in taggame API

dfabulich:


salty-horse (approved):
I'd remove the user/password option from the JSON API, as I think the basic auth should be the recommended one.

If user/pass isn't provided, should it return basic auth response headers?

dfabulich:
> I'd remove the user/password option from the JSON API, as I think the basic auth should be the recommended one.

I filed this as #1102; we can discuss it there. I'm not sure I want to remove the username/password option from the document.

> If user/pass isn't provided, should it return basic auth response headers?

The `WWW-Authenticate` is good for GET requests where authentication is required. For those cases, users can explore the API using their browser's URL bar; `WWW-Authenticate` will make the browser prompt the user to provide a username / password.

But for a POST API like this, where the user has to POST a JSON document, there's nothing that will read the `WWW-Authenticate header` and _do_ anything with it. (It's not like Python urllib is gonna pop up a username/password prompt.) 

So I'd say leave it off.

PR #1103, opened 2024-11-09
Document taggame JSON API

dfabulich:


www/api/taggame
```diff
@@ -11,73 +11,50 @@ apiPageHeader("taggame");
...
+<h2>Example request body</h2>
+
+<p>
+<pre>
+{
+    "id": "<i>the <a href="https://ifdb.org/help-tuid">TUID</a> for the game you want to update; required</i>",
+    "tags": [
+        "one tag",
+        "another tag",
+        "another tag",
```
salty-horse, line 33: "yet another tag" because there shouldn't be duplicates. Also JSON doesn't support commas after the last item.

salty-horse (approved):

Issue #1118, opened 2024-11-11
Rename patch-schema-1 and patch-schema-2


alice-blue:
@dfabulich suggested we could rename them to unscrub-ifarchive.sql and incoming-schema-changes.sql

PR #1119, opened 2024-11-11
Rename  patch-schema and patch-schema-2 to unscrub-ifarchive.sql and incoming-schema-changes.sql

alice-blue:
per suggestion from @dfabulich, to make it clearer what they are for

https://github.com/iftechfoundation/ifdb/issues/1118


alice-blue:
I guess prepare_dev_environment.sh would need to be updated as well?

dfabulich:
I've updated the `README`. "This directory" was ambiguous, so I just removed the phrase.

Issue #1050, opened 2024-11-08
Add search filters for estimated play time


dfabulich:
You should be able to filter for games where the median estimated play time is some minimum or maximum.

alice-blue:
See also #794

PR #1127, opened 2024-11-13
Search by estimated play time

alice-blue:
Allow searching <s>by "minutes:" and "hours:"</s>  to find games with an estimated play time in the given range, for example,  `playtime:30m-4h`. As part of that, this PR adds a new gametimes view, a new gametimes_mv materialized view, code to populate the materialized view, and triggers to update the materialized view. (All of that is in the upcominig schema changes file, which--I'm not sure if that's where it goes.)

This PR does not attempt to actually _display_ the estimated times in the search results--just to show results that correctly match the search terms.

Fixes https://github.com/iftechfoundation/ifdb/issues/1050

### DIFF
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -127,3 +127,97 @@ values ('bu6mmul5vxci5vqc', 'kaw2cas7dyiq2tmg', 1);
 insert into playertimes (gameid, userid, time_in_minutes)
 values ('bu6mmul5vxci5vqc', 'pwamtkqtbeyc8eyn', 6);
 
+
+
+-- The roundMedianTime function takes the exact median time in minutes
+-- and rounds it. If the time is over an hour, round to the nearest 5 minutes.
+-- Otherwise, round to the nearest minute.
+
+DELIMITER $$
+
+CREATE FUNCTION roundMedianTime(
+    exact_median_in_minutes INT(5)
+)
+RETURNS INT(5)
+DETERMINISTIC
+BEGIN
+    DECLARE rounded_median_in_minutes INT(5);
+    IF exact_median_in_minutes > 60 THEN
+        SET rounded_median_in_minutes = (round(exact_median_in_minutes/5))*5;
+    ELSE
+        SET rounded_median_in_minutes = round(exact_median_in_minutes);
+    END IF;
+    RETURN (rounded_median_in_minutes);
+END $$
+
+DELIMITER ;
+
+
+-- View to calculate the estimated play time (the rounded median time) of each game
+CREATE VIEW `gametimes` AS 
+    SELECT 
+        DISTINCT gameid, 
+        roundMedianTime( median(time_in_minutes) OVER (PARTITION BY gameid) )  as `rounded_median_time_in_minutes`
+    FROM playertimes;
+
+
+-- Create a materialized view to store the data from the gametimes view
+CREATE TABLE gametimes_mv (
+  gameid VARCHAR(32) NOT NULL,
+  rounded_median_time_in_minutes INT(5) unsigned not null,
+  PRIMARY KEY (gameid)
+);
+
+
+
+-- Populate the gametimes_mv materialized view from the gametimes view
+lock tables gametimes_mv write, gametimes read;
+truncate table gametimes_mv;
+insert into gametimes_mv select * from gametimes;
+unlock tables;
+
+
+-- Procedure to update one row of the gametimes_mv materialized view
+DROP PROCEDURE IF EXISTS refresh_gametimes_mv;
+DELIMITER $$
+
+CREATE PROCEDURE refresh_gametimes_mv (
+    IN new_gameid varchar(32) COLLATE latin1_german2_ci
+)
+BEGIN
+select *
+from gametimes
+where gameid = new_gameid
+into @gameid,
+    @rounded_median_time_in_minutes;
+if @gameid is null then
+    delete from gametimes_mv where gameid = new_gameid;
+else
+insert into gametimes_mv
+values (
+        @gameid,
+        @rounded_median_time_in_minutes
+    ) on duplicate key
+update gameid = @gameid,
+    rounded_median_time_in_minutes = @rounded_median_time_in_minutes;
+END IF;
+END;
+$$
+
+DELIMITER ;
+
+
+-- Create triggers so that when an individual player time in the playertimes table is
+-- updated, the rounded median time for that game (in the gametimes_mv materialized view)
+-- will also be updated.
+CREATE TRIGGER playertime_insert
+AFTER INSERT ON playertimes FOR EACH ROW
+call refresh_gametimes_mv(NEW.gameid);
+
+CREATE TRIGGER playertime_update
+AFTER UPDATE ON playertimes FOR EACH ROW
+call refresh_gametimes_mv(NEW.gameid);
+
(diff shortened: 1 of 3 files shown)


sql/incoming-schema-changes.sql
```diff
@@ -127,3 +127,97 @@ values ('bu6mmul5vxci5vqc', 'kaw2cas7dyiq2tmg', 1);
...
+        DISTINCT gameid, 
+        roundMedianTime( median(time_in_minutes) OVER (PARTITION BY gameid) )  as `rounded_median_time_in_minutes`
+    FROM playertimes;
+
+
+-- Create a materialized view to store the data from the gametimes view
+CREATE TABLE gametimes_mv (
+  gameid VARCHAR(32) NOT NULL,
+  rounded_median_time_in_minutes INT(5) unsigned not null,
+  PRIMARY KEY (gameid)
```
dfabulich, line 168: We should add a `KEY (rounded_median_time_in_minutes)` so we can search and sort by this.

www/searchutil.php
```diff
@@ -568,6 +570,52 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                 }
                 break;
 
+            case 'rounded_median_time_in_minutes':
+                // we need to join the gametimes mv table for this query    
+                if (!isset($extraJoins[$col])) {
+                    $extraJoins[$col] = true;
+                    $tableList .= " inner join gametimes_mv "
+                                  . "on games.id = gametimes_mv.gameid";
```
dfabulich, line 578: This means that if you search for `minutes:` with no value, no results are returned. I think we might want `LEFT OUTER JOIN` so games without time votes will return in the search results, with `rounded_median_time_in_minutes` set to `NULL`.

www/searchutil.php
```diff
@@ -568,6 +570,52 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
...
+                else if (preg_match("/^([0-9.]+)[+-]$/", $txt, $m))
+                    $expr = "$col >= '{$m[1]}'";
+                else if (preg_match("/^-([0-9.]+)$/", $txt, $m))
+                    $expr = "$col <= '{$m[1]}'";
+                else if (preg_match("/^[0-9.]+$/", $txt))
+                    $expr = "$col = '$txt'";
+                else
+                    $expr = "$col = '" . mysql_real_escape_string($txt, $db) . "'";
+                break;  
+
```
dfabulich, line 618: It's a bummer to have so much code repetition here, but I don't really have a cleverer idea. Maybe @salty-horse would have a suggestion?

dfabulich (changes requested):
I'm marking this "request changes", but the changes I'm requesting are pretty minor; we could probably merge this as-is, if you're running low on steam.

dfabulich:
Oh, uh, did we want to stick with `minutes:` and `hours:` or did we want to go with `playtime:5m-2h`?

alice-blue:
I don't know how to do the mixed hours and minutes, so if we want to do that, someone else should probably tackle it.

sql/incoming-schema-changes.sql
alice-blue: Added this.

www/searchutil.php
alice-blue: I changed it to left outer join, but wasn't sure what you meant about setting to null.

www/searchutil.php
alice-blue: I did a fall through to the next code block--you can see what you think.

alice-blue:
I think I made most of the changes you mentioned except setting to null, because I wasn't sure how to do that. However, it does seem like searching was kinda slow when I tested it just now. I'm not sure why.

www/searchutil.php
```diff
@@ -568,6 +570,30 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
                 }
                 break;
 
+            case 'rounded_median_time_in_minutes':
+            case 'rounded_median_time_in_minutes/60':
+                // we need to join the gametimes mv table for this query
```
dfabulich, line 576: If we wanted to do `3h27m`, we'd do it here. Here, we'd write a "parser" for the `$txt`.

I think the most straightforward way is to do something like this:
* `explode("-", $txt)`, and count the parts, to see if there's a min+max, or just a min, just a max, or just a straight value
* For each part, convert it into a number of minutes, with an `if`/`else if` chain, doing `preg_match` for: `/^([0-9]+)h(\d+)m$/`, then `/^([0-9]+)h$/`, then `/^([0-9]+)m$/`, then `/^([0-9+)$/` (which we'll assume to mean minutes).
* Finally, construct the `$expr` based on min/max/exact `$minutes`.

alice-blue: Do you want to do that?

dfabulich: I'm open to that, but not this week. (I was telling @salty-horse the other day that I've been spending a bit too much time on IFDB lately…)

alice-blue:
I'm attempting to do this

alice-blue:
I made some progress, but it seems like trying to enter an exact time (vs. a range) doesn't work.

www/search
```diff
@@ -910,6 +910,25 @@ else {  // ...default is searching games
...
+          searches for games with an estimated play time of 1 hour.
+
+       <p><b>hours:<i>minimum-maximum</i></b> lists games with an estimated 
+          play time in the given range. For example, <b>hours:0.5-2</b> 
+          shows games with an estimated playing time of anywhere from half an 
+          hour to 2 hours. <b>hours:4</b> lists only games with an estimated play time 
+          of 4 hours. <b>hours:6.5-</b> lists games with an estimated 
+          play time of at least 6 and a half hours. <b>hours:-8</b> lists games 
+          with an estimated play time of 8 hours or less.
+
```
dfabulich, line 931: We don't support `hours:` now, right?

alice-blue:
Hopefully that fixed it.

www/search
alice-blue: Took that out

www/searchutil.php
```diff
@@ -568,6 +594,58 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
...
+                        $expr = "$col <= '{$maximum}'";
+                    } else {
+                        // Neither minimum nor maximum time is valid, so ignore the whole thing.
+                        $expr = "";
+                    }
+                } else if (count($array_of_times) == 1) {
+                    // No hyphen was entered, so it's an exact time.
+                    $exact_time = convertTimeStringToMinutes($txt);
+                    if ($exact_time != "") {
+                        $expr = "$col == '{$exact_time}'";
```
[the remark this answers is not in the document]
alice-blue: Thanks!

www/searchutil.php
dfabulich: You did what was needed; this is the effect of `left outer join`.

```
MariaDB [ifdb]> select id, rounded_median_time_in_minutes from games inner join gametimes on games.id=gameid where title = 'Counterfeit Monkey';
Empty set (0.003 sec)

MariaDB [ifdb]> select id, rounded_median_time_in_minutes from games left outer join gametimes on games.id=gameid where title = 'Counterfeit Monkey';
+------------------+--------------------------------+
| id               | rounded_median_time_in_minutes |
+------------------+--------------------------------+
| aearuuxv83plclpl |                           NULL |
+------------------+--------------------------------+
1 row in set (0.015 sec)
```

PR #1136, opened 2024-11-15
Show tag search results in a grid

dfabulich:
@alice-blue 

I figure if it's good for https://ifdb.org/showtags to show tags in a grid, it's probably good for https://ifdb.org/search?browse&tag to do the same, and for tag search results to appear in a grid, too.

If we merge this, we might not need the `showtags` table view any more.

Open question: should we make the font smaller, like we do on `showtags`?

### DIFF
--- a/www/search
+++ b/www/search
@@ -1371,6 +1371,15 @@ else if ($term || $browse)
             }
         }
 
+        if ($searchType === "tag") {
+            global $nonce;
+            echo "<style nonce='$nonce'>\n"
+                . ".search__tags { display: grid; grid-template-columns: repeat(auto-fill, 20ch); }\n"
+                . "</style>\n"
+                . "<div class='search__tags'>\n";
+
+        }
+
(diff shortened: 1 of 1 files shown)


www/search
```diff
@@ -1371,6 +1374,15 @@ else if ($term || $browse)
             }
         }
 
+        if ($searchType === "tag") {
+            global $nonce;
+            echo "<style nonce='$nonce'>\n"
+                . ".search__tags { display: grid; grid-template-columns: repeat(auto-fill, 20ch); }\n"
```
salty-horse, line 1380: you prefer these inline instead of `ifdb.css`?

salty-horse (approved):

alice-blue:
@dfabulich I just looked at this. Between the "N games" in parentheses, the font size, and maybe also the spacing, a lot of the tag entries are wrapping onto the next line. I think it would look better without having to wrap.

Issue #1133, opened 2024-11-15
Show estimated play time in `search` API


dfabulich:
Apropos #1129:

A search for [playtime:6h](https://ifdb.org/search?searchbar=playtime%3A6h-) shows games with 6+ hours of playtime, and each search result shows the estimated play time for that game.

But if you do the same search in JSON mode <https://ifdb.org/search?searchbar=playtime%3A6h-&json> the play time doesn't appear.

PR #1145, opened 2024-11-16
Add estimated play time to search API

alice-blue:
Fixes https://github.com/iftechfoundation/ifdb/issues/1133


www/search
```diff
@@ -1088,6 +1088,10 @@ if ($api_mode) {
                 $item['coverArtLink'] =
                     get_root_url() . "coverart?id=$id&version={$row['pagevsn']}";
             }
+            // Include estimated play time if available
```
salty-horse, line 1091: The code is self-explanatory so I don't think this comment is adding anything

alice-blue: Deleted

PR #1152, opened 2024-11-18
viewgame: Add game to the "I've played it" list when you submit a time

alice-blue:
I already had this put together, and I'm not sure when we'll get to it, so I'll just submit it....

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1398,6 +1398,7 @@ function updateUnwishList(id, stat)
 <li>Vote on how long it would take to get to one final ending without hints or a walkthrough.</li>
 <li>Time spent away from the game doesn't count.</li>
 <li>Times don't need to be exact. Good-faith estimates are fine.</li>
+<li>When you vote, this game will automatically be added to your list of played games.</li>
 </ul></details>
 <p><label for='hours'>Hours:</label>
 <input type='number' name='hours' id='hours' min=0 max=200>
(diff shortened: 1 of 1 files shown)


www/viewgame
```diff
@@ -1455,7 +1452,13 @@ function submitTime() {
         document.getElementById("timeSaveMsg").innerHTML="The total time must be at least one minute.";
         return;
     }
-    // Time is valid, as far as we can tell
+    // Time is valid, as far as we can tell.
+    // If the game isn't already on the user's list of played games, add it.
+    if (document.getElementById('ckBoxckPlayed').checked==false) {
```
salty-horse, line 1457: put `document.getElementById('ckBoxckPlayed')` in a variable since it's used twice?

Also, instead of `thing.checked == false` you can do `!thing.checked`

www/viewgame
```diff
@@ -1455,7 +1451,14 @@ function submitTime() {
         document.getElementById("timeSaveMsg").innerHTML="The total time must be at least one minute.";
         return;
     }
-    // Time is valid, as far as we can tell
+    // Time is valid, as far as we can tell.
+    // If the game isn't already on the user's list of played games, add it.
+    var played_it_checkbox = document.getElementById('ckBoxckPlayed');
+    if !played_it_checkbox.checked) {
```
dfabulich, line 1457: Syntax error, missing parenthesis. (This would be a good commit to try amending!)

dfabulich (changes requested):

www/viewgame
alice-blue: I had just fixed it

PR #1153, opened 2024-11-18
Add text field for time votes: version with public usernames

alice-blue:
This is a draft. I made a start at a version that shows names publicly, but it has some problems.

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -2000,7 +2000,7 @@ function displayMedianTime($rounded_median_time, $alltimes) {
     echo "Members voted for the following times for this game:";
     echo "<ul>";
     foreach ($alltimes as $t) {
-        $time_vote_text = convertTimeToText($t['time_in_minutes']);
+        $time_vote_text = "<strong>" . convertTimeToText($t['time_in_minutes']) . "</strong>";
         $note = htmlspecialchars($t['time_note']);
         if ($t['player_time_id'] > 388) {
             // For new time votes, add the note (if present) and link to the user's profile
--- a/www/viewgame
+++ b/www/viewgame
@@ -1364,25 +1364,21 @@ function updateUnwishList(id, stat)
 //------------------PHP for the estimated play time control panel---------------------------------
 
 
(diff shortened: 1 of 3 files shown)


dfabulich:
I'm going to do some copying and pasting and make a separate PR with just public user names after a cutoff date.

www/viewgame
```diff
@@ -1454,14 +1459,17 @@ function updateUnwishList(id, stat)
...
+<li>Vote on how long it takes to get to one final ending without relying extensively on a walkthrough.</li>
 <li>Time spent away from the game doesn't count.</li>
+<li>To help players gauge how long a game with puzzles might take, you can note if your time is with or without hints.</li>
 <li>Times don't need to be exact. Good-faith estimates are fine.</li>
 </ul></details>
 <p><label for='hours'>Hours:</label>
 <input type='number' name='hours' id='hours' min=0 max=200>
 <label for='minutes'>Minutes:</label>
 <input type='number' name='minutes' id='minutes' min=0 max=59></p>
+<p><label for='timeNote'>Note:</label>
```
dfabulich, line 1471: Maybe call it a "Time Note"?

www/viewgame
```diff
@@ -1454,14 +1459,17 @@ function updateUnwishList(id, stat)
...
+<input type='text' name='timeNote' id='timeNote' maxlength='200' size='26'></p>
```
dfabulich, line 1472: Maybe add a [`placeholder`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder) attribute, like: `placeholder='e.g. "Completionist" or "No hints"'`

alice-blue:
Working on resolving this

alice-blue:
@dfabulich I tried to resolve it, but I can't get it to display the note. Maybe you can tell what I'm doing wrong.

alice-blue:
@dfabulich I think this is working now. Let me know if you want me to copy and paste everything into a new PR so it's not as messy.

PR #1156, opened 2024-11-19
Show user names next to play times, with a cutoff id

### LINKED ISSUE #1151: Make time votes public, non-anonymous
Right now, time votes are counted anonymously; there's no way to see who submitted any particular time vote.

This poses two problems:

* If we allow people to add notes to their time vote #1150, if we display them publicly, we'd be showing anonymous comments, which doesn't seem good
* If we automatically check "I've played it" when submitting a time vote #1056, but the user has set their "I've played it" list to public, we'll be making the information public.

It seems simpler and better to just make time votes public, so everyone can see who exactly submitted which votes.

dfabulich:
Before I deploy this publicly, I'll update the `player_time_id` cutoff from `4` to the current latest ID in the `playertimes` table.

Then, I'll email everyone who has submitted a time vote so far, asking them to opt in to putting their names next to the times. 

Anyone who doesn't opt in before the deadline, I'll delete their time votes, and then remove the cutoff ID

Fixes #1151

@alice-blue 

image

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -484,15 +484,12 @@ function findMedian($array_input) {
 }
 
 // Get all of the estimated play times for this game, and put them in an array
-$result = mysqli_execute_query($db, "select time_in_minutes from playertimes where gameid = ?", [$id]);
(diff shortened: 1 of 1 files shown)


salty-horse:
LGTM.

BTW, these feel more like "reports" than "votes", since users aren't picking times from a list, with the best one being shown.

salty-horse (approved):

PR #1160, opened 2024-11-19
Add time notes to public time votes

alice-blue:
@dfabulich this should be exactly the same as the other one that I said was working (https://github.com/iftechfoundation/ifdb/pull/1153 -- I copy pasted it) but with only 3 commits instead of 20. You can use whichever one you prefer.

### DIFF
--- a/www/settime
+++ b/www/settime
@@ -1,4 +1,5 @@
 <?php
+// The code in this file is for saving or deleting a time vote in the database.
 
 include_once "session-start.php";
 include_once "util.php";
@@ -21,6 +22,7 @@ if ($db == false)
 // get the request parameters
 $game = get_req_data('game');
 $newtime = get_req_data('newtime');
+$timenote = get_req_data('note');
 
 // make sure there's a game
 if ($game == "")
@@ -76,15 +78,15 @@ if ($newtime == 0 && $rid) {
 } else if ($rid) {
     // there's an existing time record, and the new time is not 0, so update the record
     $result = mysqli_execute_query($db,
-        "update playertimes set time_in_minutes = ? where id=?", [$newtime, $rid]);
+        "update playertimes set time_in_minutes = ?, time_note = ? where id=?", [$newtime, $timenote, $rid]);
     if (!$result) error_log(mysql_error($db));
     $ok = true;
 
 } else if ($newtime != 0) {
     // there's no time there yet for this user, so insert a new one
     $result = mysqli_execute_query($db,
-        "insert into playertimes (gameid, userid, time_in_minutes)
-         values (?, ?, ?)", [$game, $userid, $newtime]);
+        "insert into playertimes (gameid, userid, time_in_minutes, time_note)
+         values (?, ?, ?, ?)", [$game, $userid, $newtime, $timenote]);
     $ok = true;
 } else {
     $result = true;
--- a/www/viewgame
+++ b/www/viewgame
@@ -1521,7 +1521,7 @@ function submitTime() {
         updatePlayed();
     }
     // Submit the time.
-    jsonSend("settime?game=<?php echo $id?>&newtime=" + new_time, "timeSaveMsg", swapTimeSections);
+    jsonSend(`settime?game=<?= $id?>&newtime=${new_time}&note=${encodeURIComponent(time_note)}`, "timeSaveMsg", swapTimeSections);
 }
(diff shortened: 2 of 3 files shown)


www/viewgame
```diff
@@ -1451,15 +1453,18 @@ function updateUnwishList(id, stat)
...
+<li>To help players gauge how long a game with puzzles might take, you can note if your time is with or without hints.</li>
 <li>Times don't need to be exact. Good-faith estimates are fine.</li>
 <li>Your vote will be publicly attributed to you.</li>
 </ul></details>
 <p><label for='hours'>Hours:</label>
 <input type='number' name='hours' id='hours' min=0 max=200>
 <label for='minutes'>Minutes:</label>
 <input type='number' name='minutes' id='minutes' min=0 max=59></p>
+<p><label for='timeNote'>Time Note:</label>
+<input type='text' name='timeNote' id='timeNote' maxlength='150' size='21' placeholder='Without hints.'></p>
```
dfabulich, line 1467: This field is weird, so I like placeholder text to include some text that makes it clearly an example.

In https://github.com/iftechfoundation/ifdb/pull/1153#discussion_r1848895677 I suggested `placeholder='e.g. "Completionist" or "No hints"'`. It'll fit if we crank `size` up from 21 to 30.

Other ideas:
* `placeholder="With/without hints"`
* `placeholder="I used a walkthrough..."`

dfabulich (approved):
I think this PR is basically ready to go, (though I do have some quibbles about the placeholder text) but I want to delay shipping it for a few days, because I think we'll want to post site news about it, and I want the "public attribution" site news to be the top news item for a few days.

(We've already got ~60% of people to respond "Yes" to the email, with nobody responding "No.")

www/viewgame
dfabulich: FWIW, this is a subjective thing… people can disagree about what's good in visual design.

If we want to line up the labels for "Time Note" and "Hours", we can use a `<table>` or a CSS grid. `<table>` is easier to learn, but less flexible; CSS grid can be quite tricky to learn, but, once you learn it, it's quite handy.

alice-blue: I changed the placeholder text to say "With/without hints." and it fits. Do you still want to make the field bigger?

dfabulich: I think we're good

dfabulich:
Conflicted again

alice-blue:
Ok, it's unconflicted now.

alice-blue:
Ok, I don't know how the time_note variable went missing, but I added it in and I think it's working now.

www/viewgame
```diff
@@ -1479,14 +1484,15 @@ function updateUnwishList(id, stat)
 
 // When the user clicks the "Submit Time" button, get the input 
 // from the hours and minutes fields, validate it, convert it to 
-// minutes only, and save it to the database. 
+// minutes only, and save it to the database (with the note, if present).
 function submitTime() {
     var new_hours = document.getElementById('hours').value;
     var new_minutes = document.getElementById('minutes').value;
+    var time_note = document.getElementById('timeNote').value;
```
alice-blue, line 1491: Should I change this

`var time_note = document.getElementById('timeNote').value;`

to this?

`var time_note = document.getElementById('timeNote').value.trim();`

salty-horse: Sure :)

salty-horse: But it's good to trim stuff also in the server side before saving.

alice-blue: I wasn't sure how server-side trimming would work if encodeURIComponent() is used on line 1525. Will spaces still be trimmable after they've been converted to `%20`?

dfabulich: The server automatically unencodes URI components when you access them via `get_req_data()`. So if you type `Without hints` in the UI and it gets converted to `Without%20hints` for transmission to the server, `$timenote` will be set to `Without hints`, so trimming will work normally.

www/settime
```diff
@@ -21,6 +22,7 @@ if ($db == false)
 // get the request parameters
 $game = get_req_data('game');
 $newtime = get_req_data('newtime');
+$timenote = get_req_data('note');
```
alice-blue, line 25: So then should I change 
`$timenote = get_req_data('note');`
to this?
`$timenote = trim(get_req_data('note'));`

dfabulich: Yeah, but it's not a high priority. Separate PR?

Issue #719, opened 2021-02-23
Allow users to block any tags of their choice from appearing in news feed or search


brirush84:
This would allow users to customize their feed to, for instance, avoid games tagged 'adult'.

alyshkalia:
I'd find this feature very useful these days as so many games with AI text/art/etc. are getting uploaded to IFDB.

PR #1166, opened 2024-11-21
Filter game searches with a custom filter

alice-blue:
This is a minimal version of https://github.com/iftechfoundation/ifdb/issues/719

It allows users to create a filter that will be saved to their settings and added to their search query every time they search for games. So they can, for example, filter out games with a certain tag, or filter out games in languages they don't read, or they could require all game results to be in a certain language.

I'm not sure how I feel about it. I guess this is a draft for discussion.

To test it, go to the settings page. Type "-tag:Lovecraftian" in the new field for the game search filter, save your settings, and then do a search for "Anchorhead." That game shouldn't come up. I also tested, for example, "language:German." If you make that your filter, only games in German should show up in results.

### DIFF
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -82,7 +82,7 @@ ALTER TABLE `wishlists` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode
 -- Add column for game search filter to the users table
 --
 
-ALTER TABLE `users` ADD COLUMN `game_search_filter` VARCHAR(100) DEFAULT '';
+ALTER TABLE `users` ADD COLUMN `game_search_filter` VARCHAR(150) DEFAULT '';
 
 
 
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -76,6 +76,16 @@ ALTER TABLE `userScores_mv` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_uni
 ALTER TABLE `wishlists` CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
 
+
+
+--
+-- Add column for game search filter to the users table
+--
+
+ALTER TABLE `users` ADD COLUMN `game_search_filter` VARCHAR(100) DEFAULT '';
+
+
+
 --
 -- Table structure for table `playertimes`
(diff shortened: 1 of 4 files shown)


dfabulich (changes requested):
I feel like there should be some UI on the search results page (and the browse page!) when you apply default filters. And maybe something you can click on to redo the search with "&default_filter=0" to see all search results…?

alice-blue:
Are you talking about some kind of notification that a filter has been applied?

dfabulich:
I was imagining something a little more like the [buried-game notification](https://ifdb.org/search?searchbar=necromancer) ("Some results were hidden. [See all results](https://ifdb.org/search?searchbar=necromancer&showFlagged=1)"), but perhaps not _quite_ so in-your-face about it.

On my machine, I tried adding a default filter for `-tag:achievements` and searching for "monkey." Counterfeit Monkey is tagged "achievements," so, of course, it doesn't show. But, I claim, it's not very easy to see _why_ it's not showing.

image

I think "default filter" is a _very_ subtle notification, that likely will mean nothing to users who do a random search months or years after setting the setting.

I could imagine actually spelling out in detail exactly what additional filters applied to the current search, e.g. "Results for monkey -tag:achievements". (But maybe that would be too verbose for folks with a lot of default filters…?)

I think it would also be useful to have a clickable link on the page to redo the search without the default filters. (It would be tempting to call it "see all results" but, if the buried-game notification is showing, that might be misleading.…? 🤔)

More broadly, I think this would be a good topic to discuss on the advisors channel.

dfabulich:
I'm not sure if we reached a consensus on the [forum thread](https://intfiction.org/t/72443), but I think it'd be OK to show a constant prominent banner saying "Some results were hidden at your request. Click here to see all results."

I think that would satisfy group 1A (People who are filtering out “deal breaker” games on their own behalf) and group 2 (people filtering out something because it’s not very accessible to them).

Showing a count of how many results were hidden (or even detecting _whether_ any results were hidden) is possible, but thorny, and probably not worth the trouble. But, if you dare to try it…

* We'd first do the search not using the default custom filter, and `select` a column on whether any games would have been excluded by the custom filter.
* If no results were excluded, we'd return the search results, and not show a banner.
* If any results would have been excluded, we'd run the search _again_, this time actually excluding results, and show a banner including a count of how many results were excluded.

(I'm skeptical that it's worth the trouble to do all this!)

PR #1167, opened 2024-11-22
Remove blank.gif usage for RSS icons

salty-horse:
Also removed the unused image rss30.gif


dfabulich:
I think this had a weird effect on the vertical alignment of the RSS icon. I think I like the old vertical alignment better, but you could talk me into changing it. WDYT? 🤔 

# Before

image

# After

image

salty-horse:
I know - I think it's because images elements are aligned differently than background images.

It looks better with `bottom` instead of `center`, but only in one font size, and there are several in use. For example, `commentlog` currently looks bad [in production](https://ifdb.org/commentlog?mode=inbox), as the icon starts beneath the tops of the letters, and extends beneath the text.
commentlog


It can be tweaked individually for each font size, or maybe a separate inline element (not an `img`) will work better.

dfabulich:
Maybe we should stick with `` with `content:`? https://stackoverflow.com/a/11484688/54829

That will let us set `width` and `height` attributes, avoiding content layout shift.

```
<!doctype html>

<style>
.MyClass123{
    content:url("http://imgur.com/SZ8Cm.jpg");
}
</style>

<img class="MyClass123"/>
```

salty-horse:
What about the `commentlog` case where it isn't quite aligned correctly?

Another option is to use an SVG image that scales with the font size. ([This one](https://en.wikipedia.org/wiki/File:Feed-icon.svg) is GPL, but there might be others with a more permissive license)

salty-horse:
@dfabulich I switched to the SVG icon from the linked page (the guidelines say it's free to use in this context.)

I changed the CSS to use `em` values and a vertical position of 80%. I like how it's positioned, and I played around with the font size in the inspector to make sure it will look consistent in all situations.
rss

Issue #1117, opened 2024-11-10
Moddate on reviews seems to have gotten automatically changed when converting to UTF-8


alice-blue:
Reviews that had unusual characters are now showing the wrong moddate.

salty-horse:
The reviews table has this trigger:
```
`moddate` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp()
```

dfabulich:
I'm not sure there's anything we can/should do about this now, but maybe if we did #565 it would mitigate some of the harm.

dfabulich:
Now the reviews list their creation date correctly. They all say "* This review was last edited on November 7, 2024" but I think that's acceptable.

Issue #947, opened 2024-10-30
Inconsistent display dates for embargoed reviews


dfabulich:
When an embargoed review shows up on "New on IFDB," it displays the embargo date (today, normally). But if you click through to the review, it shows the review's creation date next to the review.

PR #1169, opened 2024-11-22
Use review creation date for display and sorting

### LINKED ISSUE #565: Review Dates Change Upon Editing
When someone updates a review, even for the most minor thing, it changes the review date. Thus, a review can be dated years off or at this point even decades off of when it was actually posted. This is confusing and bad for the historicity of the site, because you can never actually be sure when a review was posted, and it messes with the review timeline, so you can't chart the ongoing community appraisal of the work.

A better solution would be for reviews to always display the date when they were originally posted, and then maybe a small notice at the bottom of the review can note "Review last updated on x."

salty-horse:
Fixes #565, and addresses #1117.

For reviews where the edit date is a different day than the creation date, there's an "(edited: Mon, DD, YYYY)" addition. It's long, so if you have better ideas on how to display it, please share.

The personal page didn't display the date, so I removed it from the query.

Not shown, but I also checked that the RSS of game reviews is correct.

Game page:
game_page

Game page `&ratings` view:
gamepage_ratings_view

User page:
showuser

### DIFF
--- a/www/allreviews
+++ b/www/allreviews
@@ -41,8 +41,8 @@ $pg = get_req_data('pg');
 $sortReq = get_req_data('sortby');
 
 $sortList = array(
-    'new' => array('moddate desc', 'Newest First'),
-    'old' => array('moddate', 'Oldest First'),
+    'new' => array('createdate desc', 'Newest First'),
+    'old' => array('createdate', 'Oldest First'),
     'hlp' => array('netHelpful desc', 'Most Helpful First'),
     'unh' => array('netHelpful', 'Least Helpful First'),
     'hi' => array('rating desc', 'Highest Ratings First'),
@@ -195,12 +195,12 @@ if ($errMsg) {
            reviews.rating as rating,
            reviews.summary as summary,
            reviews.review as review,
-           reviews.moddate as moddate,
+           reviews.createdate as createdate,
            if (reviews.embargodate > now(),
              date_format(reviews.embargodate, '%M %e, %Y'),
              null) as embargodate,
            reviews.special as special,
-           date_format(reviews.moddate, '%M %e, %Y') as moddatefmt,
+           date_format(reviews.createdate, '%M %e, %Y') as createdatefmt,
            sum(reviewvotes.vote = 'Y' and ifnull(rvu.sandbox, 0) in $sandbox) as helpful,
            sum(reviewvotes.vote = 'N' and ifnull(rvu.sandbox, 0) in $sandbox) as unhelpful,
            ifnull(sum(reviewvotes.vote = 'Y'), 0)
--- a/www/game-rss.php
+++ b/www/game-rss.php
@@ -42,7 +42,7 @@ function getGameRssItems($db, $id, $feedType, $gameTitle, $links, $extFeed)
 
         // query the reviews
         $result = mysql_query(
-            "$selectMemberReviews order by moddate desc", $db);
+            "$selectMemberReviews order by createdate desc", $db);
 
         // build the item
         for ($i = 0 ; $i < mysql_num_rows($result) ; $i++) {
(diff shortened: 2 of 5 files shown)


dfabulich:
There as many as _three_ dates for each review: its creation date, its embargo date, and the edit date (if any). #947 

In `newitems.php` we order by `greatest(reviews.createdate, ifnull(reviews.embargodate, '0000-00-00')) as d` which feels like the right thing to do.

> For reviews where the edit date is a different day than the creation date, there's an "(edited: Mon, DD, YYYY)" addition. It's long, so if you have better ideas on how to display it, please share.

I like the suggestion in #565 to add a note in a smaller font to the bottom of the review, giving us room to make the note as long/complicated as needed, and to display the computed sorting date at the top of the review.

> _This review was originally created on XXX, and was embargoed until YYY. It was last edited on ZZZ._

> _This review was originally created on XXX, and was last edited on YYY._

dfabulich (changes requested):
There as many as _three_ dates for each review: its creation date, its embargo date, and the edit date (if any). #947 

In `newitems.php` we order by `greatest(reviews.createdate, ifnull(reviews.embargodate, '0000-00-00')) as d` which feels like the right thing to do.

> For reviews where the edit date is a different day than the creation date, there's an "(edited: Mon, DD, YYYY)" addition. It's long, so if you have better ideas on how to display it, please share.

I like the suggestion in #565 to add a note in a smaller font to the bottom of the review, giving us room to make the note as long/complicated as needed, and to display the computed sorting date at the top of the review.

> _This review was originally created on XXX, and was embargoed until YYY. It was last edited on ZZZ._

> _This review was originally created on XXX, and was last edited on YYY._

salty-horse:
Your suggestion includes creation date twice, at the top and bottom. Isn't that redundant?

I still think there should be an indication something was edited. Maybe an asterisk next to the top date?

Should the embargo date be public knowledge?

dfabulich:
> I still think there should be an indication something was edited. Maybe an asterisk next to the top date?

Sure, an asterisk would be fine.

> Should the embargo date be public knowledge?

I don't see why not. It's certainly publicly available in the IF Archive export.

> Your suggestion includes creation date twice, at the top and bottom. Isn't that redundant?

Kinda? I think the explanation would be confusing unless it juxtaposes the dates, which inherently requires repeating whichever date happens to be at the top.

For example, suppose we have three reviews:

1. Created Jan 1, embargoed until Jan 2, edited Jan 3
    * At the top: "Jan 2*"
    * At the bottom: "* This review was originally created on Jan 1, and it was embargoed until Jan 2. It was last edited on Jan 3."
2. Created Feb 1, embargoed until Feb 2
    * At the top: "Feb 2*"
    * At the bottom: "* This review was originally created on Feb 1, and it was embargoed until Feb 2."
3. Created Mar 1, edited Mar 3
    * At the top: "Mar 1*"
    * At the bottom: "* This review was originally created on Mar 1. It was last edited until Mar 3."

dfabulich:
OK, I think the consensus from the [forum thread](https://intfiction.org/t/72503) is to handle it like this:

* The main date at the top should be the "publication date," which is the greater of either the creation date or the embargo date
* If the review was edited, show the date with an asterisk, and then, at the end, show "* This review was last edited on [DATE]." (We don't need to repeat the publication date, or separately specify the creation date or the embargo date.)

salty-horse:
What about "rating only" reviews, that are just a rating, username, and date? (Shown in a screenshot in the first post)

salty-horse:
Pushed a new commit.

For big reviews, I used a footnote (with `<small>` element rather than `<div class="details">`), and added a title to the asterisk with the same text.

For ratings-only reviews, I left it on the same line as the creation date.
edited

Issue #1155, opened 2024-11-19
Time votes don't have a created date or a modified date


dfabulich:
```
MariaDB [ifdb]> describe playertimes;
+-----------------+-----------------+------+-----+---------+----------------+
| Field           | Type            | Null | Key | Default | Extra          |
+-----------------+-----------------+------+-----+---------+----------------+
| id              | int(11)         | NO   | PRI | NULL    | auto_increment |
| gameid          | varchar(32)     | NO   | MUL | NULL    |                |
| userid          | varchar(32)     | NO   |     | NULL    |                |
| time_in_minutes | int(5) unsigned | NO   |     | NULL    |                |
+-----------------+-----------------+------+-----+---------+----------------+
```

It's a good practice to have a `created` date as well as a `modified` date.

I was going to use the vote `created` date as a way of transitionally implementing #1151 but, uh, I guess I'm not gonna do _that_.

alice-blue:
Since they're not very modifiable, do we still need a modifed date?

dfabulich:
I think one `date` column would be a big improvement.

alice-blue:
Create date has been added.

PR #1170, opened 2024-11-22
Add create date to player times

alice-blue:
Add a create date column to the playertimes table, and automatically insert the date when a time is added.

(Only the last two commits relate to the create date. All the previous commits are from the time note PR. Which might not be the right way to do it, but I was trying to avoid more conflicts.)

This addresses https://github.com/iftechfoundation/ifdb/issues/1155


alice-blue:
Now that time notes are live, I'll make a separate PR with just the createdate commits.

Issue #558, opened 2021-11-20
Suggest displaying top ten reviewers instead of top four


aubloom:
The amount of points now required to reach the site's front page top four is enormous. I say this as someone sitting at position four as I type this. I feel like it might reintroduce some healthy competition, and make getting on the front page seem more acheivable (rather than like that none of the people there can ever be passed) if the list was expanded to show ten people. The change itself might spur some action/thought, and simultaneously, people in the top four aren't punished at all for the work they've already put in. Sure, in the LONG run, the same problem could happen to the top ten. But none of us live forever. I think my idea's pretty good for now.

Issue #776, opened 2021-02-17
Hot Reviewers: Top reviewers in the last N weeks/months


qdacsvx:
A chart of ?5 reviewers who published the best (or the most!) reviews in the past week. Mathbrush would still do well, and rightly so. But there could be new faces each week depending who is having a go. There is an incentive to get reviewing because you could be listed too.

Should this replace Top Reviewers?

dfabulich:
I started working on this, but since `reviewvotes` only gained a `createdate` column yesterday, _all_ of the votes are new within the last month.

Still, this query did something useful. In early 2025 I can try it again, changing the interval from `12 hour` to `30 day`.

<details><summary>Query</summary>
<p>

```
select `recentuserscoreitems`.`userid` AS `userid`,
  sum(`recentuserscoreitems`.`score`) AS `score`,
  (
    max(`recentuserscoreitems`.`isReview`) * sum(`recentuserscoreitems`.`score`)
  ) AS `rankingScore`,
  sum(`recentuserscoreitems`.`isReview`) AS `reviewCount`
from (select `reviews`.`userid` AS `userid`,
(
    max(if(isnull(`reviews`.`review`), 10, 100)) + (
      5 * greatest(
        -(100),
        least(
          100,
(
            ifnull(
              sum(
                (
                  `reviewvotes`.`vote` = 'Y'
                  and ifnull(`users`.`sandbox`, 0) = 0
                )
              ),
              0
            ) - ifnull(
              sum(
                (
                  `reviewvotes`.`vote` = 'N'
                  and ifnull(`users`.`sandbox`, 0) = 0
                )
              ),
              0
            )
          )
        )
      )
    )
  ) AS `score`,
  max(if(isnull(`reviews`.`review`), 0, 1)) AS `isReview`
from (
    (
      `reviews`
      left join `reviewvotes` on((`reviewvotes`.`reviewid` = `reviews`.`id`))
      left join `users` on ((`users`.`id` = `reviewvotes`.`userid`))
    )
    left join `specialreviewers` on((`reviews`.`special` = `specialreviewers`.`id`))
  )
where (
    (
      isnull(`reviews`.`special`)
      or (not(`specialreviewers`.`editorial`))
    )
    and ifnull((now() >= `reviews`.`embargodate`), 1)
    and reviewvotes.createdate > date_sub(now(), interval 12 hour)
  )
group by `reviews`.`id`
union all
select `reviewvotes`.`userid` AS `userid`,
[...]

PR #1171, opened 2024-11-22
Show top 10 instead of top 4 reviewers

alice-blue:
Assuming people are on board with this suggestion: https://github.com/iftechfoundation/ifdb/issues/558

Before:

top 4 reviewers

After:

top 10 reviewers


dfabulich:
I guess this is OK for now, but I think we'll probably want to crank it back down again if/when we get around to showing the top reviewers of the month. #776

Issue #1080, opened 2024-11-08
Show your own time votes in your "I've played this" list


alice-blue:
On this page:

https://ifdb.org/playlist?type=played

The playlist page already shows the user's star ratings. Maybe it can show the time votes, too, so that users can easily see which games they've added times for.

alice-blue:
Maybe this should be done as part of https://github.com/iftechfoundation/ifdb/issues/582

alice-blue:
Now that time votes are public, this could be showed on all played lists, not just your own.

alice-blue:
Ok, I have a branch to show times on the "I've played it" list, but I'm not sure whether to create a PR. Right now I have it set up so that it will display time votes regardless of whether it's your own played list or not, but it will only display the non-anonymous ones (the ones above record number 388). Even on your own played list. Is that confusing? Should I just wait until Dec. 19 for this?

PR #1172, opened 2024-11-23
Show the user's play time on the "I've played it" list

alice-blue:
Fixes https://github.com/iftechfoundation/ifdb/issues/1080

This version shows public time votes to everyone who can view the "played" list. It shows anonymous time votes only if the list belongs to the current user, and adds a note saying the time is anonymous until December 19.

After Dec. 19, some of the code for dealing with anonymous votes can be removed.

I mostly tested looking at the user's own list. I didn't test looking at a list from the perspective of another user.

### DIFF
--- a/www/playlist
+++ b/www/playlist
@@ -87,7 +87,14 @@ if ($listtype == "reviewideas") {
     $listTable = "playedgames";
     $listName = "play list";
     $title = trim(htmlspecialcharx("$usernamePoss Played Games"));
+    $listCols = ", playertimes.time_in_minutes, playertimes.id as time_id";
     $privcode = "P";
+    $listJoin = "join playedgames"
+                .  " on playedgames.gameid = games.id"
+                .  " and playedgames.userid = '$quid'"
+                .  " left join playertimes"
+                .  " on playedgames.gameid = playertimes.gameid"
+                .  " and playertimes.userid = '$quid'"; 
 }
 
 if ($listTable && !$listJoin) {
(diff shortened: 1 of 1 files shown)


www/playlist
```diff
@@ -234,6 +241,17 @@ if ($errMsg) {
...
             $onPlayList = ($listtype == "reviewideas" && $rec['onPlayList']);
+            $user_time = "";
+            if ($rec['time_in_minutes'] && $rec['time_id']) {
+                if ($rec['time_id'] > 388) {      
+                    // If the time is public
+                    $user_time = convertTimeToText($rec['time_in_minutes']);
+                } else if ($uid == $curuser) {
+                    // If the time is anonymous, and belongs to the current user
+                    $user_time = convertTimeToText($rec['time_in_minutes']);
+                    $user_time .= " (anonymous <a href='https://ifdb.org/news?item=152'>until December 19</a>)";
```
dfabulich, line 252: Rather than add another place where we're checking the time ID, I think I want to wait on merging this PR until after Dec 19. That way, we won't have to remember to rip it out again in a few weeks.

alice-blue: Ok.

alice-blue:
I know we don't want to merge this until (after?) Dec. 19, but I removed the conditional code for dealing with anonymous votes, so this can be reviewed whenever. I tested by adding times as the Test Tester user, looking at my playlist, and then by logging out and looking at Test Tester's playlist by going to /playlist?id=0000000000000000. I think.

PR #1175, opened 2024-11-23
Search page cleanup

salty-horse:
* This touches other pages, too: **Use `<aside>` for tipboxes, add a title to the 'create style' tipbox**
* **Remove duplication in search URL query parsing** - the two loops set the same variables, so I removed the duplication, and turned the map into a const.
* **Fix undefined variable in /search?game** - In `helpExtLink()`, I think this was a copy-paste error from `helpListLink()` that needs the `$id` to expand the section when clicking it.
* Remove unused `newSearch` function - it was removed in d8181466fbd178c01f1291c3537bfaa2f5efaa78
* Move search form into a function - I think I got everything right. The search box in the bottom sets an `id`, but it's unused. For some reason, the game search form included all the help text, and I don't think there was any hidden input element there, so I moved it outside, and unindented it (View the commit with "ignore whitespace" to see the actual changes)

### DIFF
--- a/www/search
+++ b/www/search
@@ -1275,58 +1275,58 @@ else if ($term || $browse)
         // if this is a game search, show the tip boxes
         if ($searchType == "game") {
             echo "<div class=tipgroup>"
-                . "<div class=tipbox>"
+                . "<aside class=tipbox>"
                 . "<h1><a href=\"search\">Search tips</a></h1>"
                 . "Learn more about search syntax."
-                . "</div>"
-                . "<div class=tipbox>"
+                . "</aside>"
+                . "<aside class=tipbox>"
                 . "<h1><a href=\"editgame?id=new\">Add a game listing</a></h1>"
                 . "You can help improve IFDB by adding a game to our database."
-                . "</div>";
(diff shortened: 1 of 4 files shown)


www/search
```diff
@@ -222,18 +222,7 @@ $qterm = mysql_real_escape_string($term, $db);
...
-$searchButton = "Search Games";
-foreach (SEARCH_TYPES as $st) {
-    if (isset($_REQUEST[$st])) {
-        $searchType = $st;
-        break;
-    }
-}
-
-$browseTabs = [
+const BROWSE_TABS = [
```
dfabulich, line 225: I feel like this would be better named `SEARCH_TYPES` than `BROWSE_TABS`. In the current implementation, it's mostly used for types of searches.

dfabulich (changes requested):

salty-horse:
Renamed to `SEARCH_TYPES`. Once again, I force-pushed and reflexively rebased onto the latest `main`, so the "Compare" button offered by Github will show more changes than intended.

PR #1177, opened 2024-11-23
Add a create date to the playertimes table

alice-blue:
Add a create date column to the playertimes table, and automatically insert the date when a time is added.

(This PR includes only the two commits related to the create date.)

This addresses https://github.com/iftechfoundation/ifdb/issues/1155

### DIFF
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -87,6 +87,7 @@ CREATE TABLE playertimes (
   userid VARCHAR(32) NOT NULL,
   time_in_minutes INT(5) unsigned not null,
   time_note VARCHAR(150),
+  createdate DATE,
   PRIMARY KEY (id),
   UNIQUE KEY `game_user` (`gameid`, `userid`)
 );
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -87,7 +87,7 @@ CREATE TABLE playertimes (
   userid VARCHAR(32) NOT NULL,
   time_in_minutes INT(5) unsigned not null,
   time_note VARCHAR(150),
-  createdate DATE DEFAULT (CURRENT_DATE),
+  createdate DATETIME DEFAULT now(),
   PRIMARY KEY (id),
   UNIQUE KEY `game_user` (`gameid`, `userid`)
 );
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -87,7 +87,7 @@ CREATE TABLE playertimes (
   userid VARCHAR(32) NOT NULL,
   time_in_minutes INT(5) unsigned not null,
   time_note VARCHAR(150),
-  createdate DATE,
+  createdate DATE DEFAULT (CURRENT_DATE),
   PRIMARY KEY (id),
   UNIQUE KEY `game_user` (`gameid`, `userid`)
 );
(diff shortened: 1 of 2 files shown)


sql/incoming-schema-changes.sql
```diff
@@ -87,6 +87,7 @@ CREATE TABLE playertimes (
   userid VARCHAR(32) NOT NULL,
   time_in_minutes INT(5) unsigned not null,
   time_note VARCHAR(150),
+  createdate DATE,
```
salty-horse, line 90: I think `createdate DATE DEFAULT (CURRENT_DATE)` will make sure it's set automatically, so you even don't need to mention it when inserting a row.

(I haven't tested this)

alice-blue:
I was wondering what to do about all of the existing rows that won't have a date. Do we need to do something like this?

```
UPDATE playertimes
SET createdate = '2014-11-23';
```

sql/incoming-schema-changes.sql
alice-blue: [Stack Overflow says](https://stackoverflow.com/questions/20461030/current-date-curdate-not-working-as-default-date-value/51709343#51709343)

> According to this [documentation](https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html), starting in MySQL 8.0.13, you will be able to specify:
> 
> CREATE TABLE INVOICE(
>     INVOICEDATE DATE DEFAULT (CURRENT_DATE)
> )

I don't know what version we are using, but it sounds promising.

alice-blue: Ok, I changed this.

dfabulich:
The new default value will take care of it automatically.

sql/incoming-schema-changes.sql
```diff
@@ -87,6 +87,7 @@ CREATE TABLE playertimes (
   userid VARCHAR(32) NOT NULL,
   time_in_minutes INT(5) unsigned not null,
   time_note VARCHAR(150),
+  createdate DATE DEFAULT (CURRENT_DATE),
```
dfabulich, line 90: (Sorry, I thought I'd posted about this before, but I guess I didn't.)

```suggestion
  createdate DATETIME DEFAULT now(),
```

> https://stackoverflow.com/questions/31761047/what-difference-between-the-date-time-datetime-and-timestamp-types
> 
> **DATE:** It is used for values with a date part but no time part. MySQL retrieves and displays DATE values in ***YYYY-MM-DD*** format. The supported range is `1000-01-01` to `9999-12-31`.
> 
> **DATETIME:** It is used for values that contain both date and time parts. MySQL retrieves and displays DATETIME values in ***YYYY-MM-DD HH:MM:SS*** format. The supported range is `1000-01-01 00:00:00` to `9999-12-31 23:59:59`.
> 
> **TIMESTAMP:** It is also used for values that contain both date and time parts, and includes the time zone. TIMESTAMP has a range of `1970-01-01 00:00:01` UTC to `2038-01-19 03:14:07` UTC.

dfabulich (changes requested):

sql/incoming-schema-changes.sql
alice-blue: What would we use the time for?

dfabulich: Sorting, at least. I would have used it for the deanonymization thingy.

alice-blue: Done

PR #1180, opened 2024-11-25
Improve Personal page markup

salty-horse:
The `personal` page used table layout for its main content and tip boxes.

* The tip boxes used a style unique to the page, despite there being tip boxes elsewhere. I changed those to the common style.
* I also updated the page with `<main>`, `<aside>` and `<section>` elements with `<h1>`. (Tested in dark mode)
* The "play list" in the "Your Play Lists" section was missing a headline - I added it.

One thing I'm unsure about is the width of the "main" content. In the table layout, it filled the part of the page that wasn't taken up by the tip boxes.

Now I let it be as wide as it wants to be, but wrapped the help messages when sections are empty with `<p>`, so they will get the `var(--paragraph-width)` style, but it seems too thin.

I think `ifdb.css` has a misleading comment about it:
```css
:root {
    /* The width of the main sections of a page, without any sidebars. */
    --paragraph-width: 60ch;
}
```
This is definitely NOT the "width of the main sections of a page, without any sidebars". It's only applied to a few elements (`p, ul, ol`), and most pages have text that extends beyond it. (See the news in the home page, or the headlines of reviews in a game page)

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -1484,18 +1484,6 @@ div.cpanelHeadingTop {
 
 
 /*
-The "boxhead" style is used for the headline within "rightbar"
-boxes.
-*/
-tr.boxhead {
-    margin: 0 0 0 0;
-    padding: 0 0 0 0;
-}
-tr.boxhead td {
-    text-align: left;
-}
-
-/*
 The "dots" is used for horizontal rules that separate sections
 on a page.
 */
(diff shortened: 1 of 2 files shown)


dfabulich:
This PR is conflicted now, but also, it's _semantically_ conflicted by #1175, as tipboxes now have to be `<aside>` elements instead of `<div>` elements.

> One thing I'm unsure about is the width of the "main" content

On my monitor, the header backgrounds are extending behind the tip boxes, which feels wrong. It feels a bit better to remove `class=headline` and just display the headings as big bold H1s.

image

image

dfabulich (changes requested):
Conflicted

salty-horse:
Fixed conflicts, removed `h1.headline` class.

Issue #1034, opened 2024-11-08
Set time vote controls to "display:none" in the CSS so that they won't appear unless javascript is enabled


alice-blue:
(Edited to try to clarify what I am suggesting)

The time vote controls rely on javascript to hide and show the correct sections. We should test what happens if javascript is not enabled, and maybe hide the sections by default in the CSS, since it would look very confusing to show them all at the same time. 

And maybe there needs to be a message there that JS is required?

Users with javascript enabled would still see the time vote controls, because there is already javascript in place to display the correct section. If they have javascript enabled, that code will run.

salty-horse:
Since they give an error if you haven't played the game, they should only be visible if that's true.

dfabulich:
I think the thought process behind #383 should apply here: if you submit a time vote, we should check the "I've played it" box for you automatically. And so, IMO, the time-vote form should be _visible_ by default.

alice-blue:
I think maybe my suggestion didn't come across clearly. Let me try again:

I'm not suggesting we change the experience of users who have javascript enabled. This suggestion would only affect the experience of users who do not have javascript enabled. 

The idea is that we set the divs to "display:none" in the CSS. The time controls will still be visible to users who have javascript enabled, because there's already javascript in place to dispay them after the page loads. But they won't be visible to people who do not have javascript enabled, because the javascript to display the correct div will not run. For those users, there could be a message that javascript is required.

PR #1182, opened 2024-12-01
Hide time vote controls when JavaScript is disabled

alice-blue:
The time vote controls rely on JS to hide and show the correct sections. This PR hides the sections by default in the CSS, because otherwise, if JS is disabled, all the controls will show simultaneously in a way that doesn't make sense.

This also adds a message that JS is required.

Fixes https://github.com/iftechfoundation/ifdb/issues/1034

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -657,6 +657,20 @@ the page.
 }
 
 /*
+The time vote controls in the "Have you played this game?" box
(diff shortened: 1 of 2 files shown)


salty-horse:
Maybe the page should refresh when submitting a votes, like a classic POST request.

PR #1183, opened 2024-12-01
Create a custom filter for game searches -- Revised version

alice-blue:
This is a revised version of #1166.

Per suggestions on the forum, it has a link at the bottom of the results page saying that results were filtered, and to click to search again without the filter. The message follows the font size of nearby text (either the same size as the page controls when there are multiple search results, or the same size--and italics--as "no results found" when there are no search results).

I also revised the description on the settings page and tried to make it clear that this is not a reliable way to ensure kid-friendly content.

I also increased the max length of the filter to 150 characters.

I'm not positive the "search again" link works the way it is supposed to. I was guessing at how to make it work, and don't know if there are other variables or options that need to be taken into account when redoing the search.

This addresses part of https://github.com/iftechfoundation/ifdb/issues/719. I haven't figured out how to do the rest of it.


dfabulich:
This PR is conflicted now that we're using the Dec 1 IF Archive data drop. (There are no "incoming" schema changes now that the archive includes all of the schema changes prior to Dec 1.) You can unconflict it just by copying and pasting your schema change to the new version (nearly blank) version of `incoming-schema-changes.sql`.

alice-blue:
Ok, thanks. I think I fixed it.

www/search
```diff
@@ -1226,6 +1227,10 @@ else if ($term || $browse)
 
         echo "<i>No results were found.</i>";
 
+        if ($games_filtered == true) {
+            echo '<p><i>A filter was applied to your search. <a href="search?searchfor=' . $term . '&nofilter=1">Click</a> to search again without the filter.</i></p>';
```
dfabulich, line 1231: This breaks when using `/search?browse`. In that case, there are no terms, and clicking the link takes you to `/search?searchfor=&nofilter=1` without browse mode, showing you the Advanced Search Tips instead of unfiltered browse results.

dfabulich (changes requested):

www/search
```diff
@@ -1226,6 +1227,10 @@ else if ($term || $browse)
 
         echo "<i>No results were found.</i>";
 
+        if ($games_filtered == true) {
```
dfabulich, line 1230: It's marginally better to say `if ($games_filtered)` without `== true`

www/search
```diff
@@ -1663,6 +1668,11 @@ else if ($term || $browse)
         // add the page controls again at the bottom, if applicable
         if ($pg != 1 || $rowcnt > $perPage)
             echo "<div class='search__pageCtl'>$pageCtl</div>";
+
+        // Announce that the game results are filtered, if applicable
+        if ($games_filtered == true) {
```
dfabulich, line 1673: here, too, no need for `== true`

alice-blue: Changed this

www/search
alice-blue: Ok, I tried to fix this.

www/search
alice-blue: Changed this.

dfabulich (changes requested):
This is now broken by changes in #1191. This PR adds a new parameter to `doSearch()`, `$override_game_filter`, but now we also have to add that to `ifdb-recommends.php`.

alice-blue:
I added $override_game_filter and $games_filtered to ifdb recommends. (I don't really know what to do with all the values that don't get used.)

dfabulich:
I caught a bad bug: add a custom filter to your settings, then search Recommended Lists for `author:mathbrush`. http://localhost:8080/search?searchfor=author%3Amathbrush&searchgo=Search+Lists&list=1

> No results were found.
>
> Game results were filtered at your request. [Search again without the filter.](http://localhost:8080/search?searchfor=author:mathbrush&nofilter=1)

I assume it's trying to apply the filter to non-game searches or something…?

image

alice-blue:
Tried to fix this.

alice-blue:
I opened a new PR that doesn't have the strange formatting, so I'm closing this one.

PR #1186, opened 2024-12-02
ifdb-recommends.php: Delete legacy code

alice-blue:
Delete automatic recommendation algorithm that finds users who are “similar” to you and finds games that they recommend highly


www/components/ifdb-recommends.php
```diff
@@ -4,293 +4,6 @@
...
-          }
-          xhr.send();
-        }();
-      </script>
-      <?php
-      return;
-    }
-}
 
 // if our recommendation list is empty, pick some random 4- and 5-star
```
dfabulich, line 8: Let's fix this comment and remove the `if` statement

dfabulich (changes requested):
Let's remove `$overloaded`, `$recsrc`, and `$_SESSION`, too.

www/components/ifdb-recommends.php
alice-blue: Ok. Does that look right?

www/components/ifdb-recommends.php
```diff
@@ -3,300 +3,9 @@
...
-}
-
-// if our recommendation list is empty, pick some random 4- and 5-star
-// games as recommendations
-if (count($recs) == 0) {
-
-    if ($debugflag) echo "falling back on default method<br>";
+// Pick some random 4- and 5-star games as recommendations
 
     // if we're logged in, don't pick games we've already rated
```
dfabulich, line 10: Thanks! Let's also remove the indentation that was here for the `if` block.

dfabulich (changes requested):

www/components/ifdb-recommends.php
alice-blue: Ok, I think I got it.

PR #1189, opened 2024-12-02
newitems: Don't show Details link on short news items

dfabulich:
Try creating a very short competition news item like this one: https://ifdb.org/newslog?newsid=608

The home page (or <https://ifdb.org/allnew>) will display it with a "- Details" link, but all the important information is already visible, and the news item itself already links to the competition (or game, if there's game news); there's no need for a link.

### DIFF
--- a/www/newitems.php
+++ b/www/newitems.php
@@ -633,7 +633,7 @@ function showNewItemList($db, $first, $last, $items, $options)
             $nuidOrig = $n['origUserID'];
             $nunameOrig = htmlspecialcharx($n['origUserName']);
             $nhead = htmlspecialcharx($n['headline']);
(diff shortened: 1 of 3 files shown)


salty-horse:
The link itself is useful - it's a permalink. Maybe the problem is just with its name or location?

dfabulich:
I think the issue is its location, when it provides no additional detail. It lures the user into clicking on it, expecting more details, but none will be provideded.

I added a commit to add a permalink to all the places where we show news (`newslog`, `viewgame` `viewcomp`). Anyone who wants a permalink can now access it via the link to the competition, expand the top news item, and access the "direct link" there, if they wish.

salty-horse (approved):

PR #1191, opened 2024-12-02
Use searchutil for ifdb recommends

alice-blue:


www/searchutil.php
```diff
@@ -257,6 +257,7 @@ function doSearch($db, $term, $searchType, $sortby, $limit, $browse)
         $selectList = "distinct games.id as id,
                        games.title as title,
                        games.author as author,
+                       games.desc as description,  // On the home page, descriptions are shown for recommended games
```
dfabulich, line 260: `//` comments don't work inside quotation marks. This breaks the search.

dfabulich (changes requested):

www/searchutil.php
alice-blue: Oops. I can take it out.

alice-blue: Done

PR #1196, opened 2024-12-04
Use the doSearch function for getting new games on the front page

alice-blue:


alice-blue:
Oh--I guess `"limit 0, $games_limit"` could just be  `"limit $games_limit"` instead? I did it with the 0 because I saw it done that way somewhere else (IFDB Recommends?).

dfabulich:
This PR is doing a full scan of the games table, but I confirmed that if I merge in #1265, it's doing a fast index lookup.

Let's merge #1265, and then merge #1196

PR #1199, opened 2024-12-04
Filter games version 3

alice-blue:
Slightly more verbose message. Fixed the spacing before the message (by using `<p>`) so that the spacing is about the same whether there is one result or multiple results. Changed some variable names to be clearer (taking out "search" since the game results aren't always the result of obvious searches). Also, the formatting doesn't look so strange anymore (I never figured out the cause).  There's now a default value for the $override_game_filter parameter (though I don't know if it'll help much).

### DIFF
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -7,3 +7,7 @@ alter table reviewvotes
     add column `createdate` datetime NOT NULL DEFAULT current_timestamp(),
     add PRIMARY KEY (`reviewvoteid`)
 ;
+
+-- Add column for game search filter to the users table
+
+ALTER TABLE `users` ADD COLUMN `game_filter` VARCHAR(150) DEFAULT '';
--- a/www/components/ifdb-recommends.php
+++ b/www/components/ifdb-recommends.php
@@ -21,12 +21,13 @@ $sortby = "ratu";  // Sort the highly rated games to the top of the results.
 $maxpicks = 12;    // Get the first twelve results. (We want extras so we're not always displaying the same games.) 
 $limit = "limit 0, $maxpicks";
 $browse = 0;
+$override_game_filter = 0;
 
 
 // run the search for highly-rated games
 list($recs, $rowcnt, $sortList, $errMsg, $summaryDesc, $badges,
-    $specials, $specialsUsed, $orderBy) =
-    doSearch($db, $term, $searchType, $sortby, $limit, $browse);
+    $specials, $specialsUsed, $orderBy, $games_were_filtered) =
+    doSearch($db, $term, $searchType, $sortby, $limit, $browse, $override_game_filter);
 
 
 // show some recommendations
--- a/www/editprofile
+++ b/www/editprofile
@@ -56,6 +56,7 @@ $emailcloaked = (get_req_data('emailcloaked') != 0);
 $gender = get_req_data('gender');
 $offsite = get_req_data('offsite');
 $accessibility = (get_req_data('accessibility') ? 1 : 0);
+$game_filter = get_req_data('game_filter');
 
 if (strlen($gender) != 1 || strpos("MF", $gender) === false)
     $gender = "";
@@ -72,7 +73,7 @@ function showHiddenFields()
     global $email, $email2, $pubemail, $dname, $loc, $profile, $pic,
         $defaultOS, $mirror, $noExes,
         $playlistPub, $wishlistPub, $unwishlistPub, $gender,
-        $emailcaptcha, $emailcloaked;
+        $emailcaptcha, $emailcloaked, $game_filter;
 
     echo "<input type=hidden name=email value=\""
           . htmlspecialcharx($email) . "\">"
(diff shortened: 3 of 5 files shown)


alice-blue:
Here's a screenshot of "no results found." I didn't get a screenshot of the other one.

game filtering v3 A

alice-blue:
Revised it again.

filter message 3

www/newitems.php
```diff
@@ -167,6 +174,36 @@ function getNewItems($db, $limit, $itemTypes = NEWITEMS_ALLITEMS, $options = [])
...
+        if ($curuser && $override_game_filter != 1) {
+            $result = mysqli_execute_query($db, "select game_filter from users where id = ?", [$curuser]);
+            if (!$result) throw new Exception("Error: " . mysqli_error($db));
+            [$game_filter] = mysql_fetch_row($result);
+            if ($game_filter != "") {
+                // Find games that have at least one review, and use the custom game filter to filter them
+                $term = "#reviews:1-";
+                $searchType = "game";
+                $sortby = "lnew";
+                $limit = null;
```
dfabulich, line 190: With `$limit = null`, this returns all 6000+ games that have reviews. The query plan seems to indicate that it's doing a full scan of the `games` table.

```
+------+--------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------------------------------------------+
| id   | select_type  | table                  | type   | possible_keys | key     | key_len | ref           | rows  | Extra                                        |
+------+--------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------------------------------------------+
|    1 | PRIMARY      | games                  | ALL    | NULL          | NULL    | NULL    | NULL          | 13984 | Using where; Using temporary; Using filesort |
|    1 | PRIMARY      | gameRatingsSandbox0_mv | eq_ref | PRIMARY       | PRIMARY | 130     | ifdb.games.id | 1     |                                              |
|    2 | MATERIALIZED | gametags               | ref    | gameid,tag    | tag     | 1002    | const         | 351   | Using where                                  |
+------+--------------+------------------------+--------+---------------+---------+---------+---------------+-------+----------------------------------------------+
```

It seems to add 0.2 seconds to loading the home page on my laptop, which I guess is fine if hardly anybody uses this feature, but it creeps me out that if this feature gets even a little bit popular, it could take down the site.

I dunno, maybe it's fine. #1203 was fine for a long time, until it wasn't, and it took down the site for a few minutes. But the site came back up again on its own. And then we fixed it, and now it's fine again.

I think the "right way" is to do what I suggested here:

> I guess we might have to start by adding a feature to sort games by "Most Recently Reviewed." To do _that_ in a performant way, we'd need to use a materialized view. I… think it might be possible to add a `lastReviewDate` column to `gameRatingsSandbox0_mv`…? But that would mean wading in to the nightmarishly complicated `gameRatingsSandbox0` query in `unscrub-ifarchive.sql`.

I'd feel a lot better about merging this PR if we did that. But maybe I just need to get over myself and merge it?

(Sorry I've left this hanging for a month.)

Maybe @salty-horse has opinions?

dfabulich:
Oh, also, it's conflicted, but perhaps the deeper problem is that _I'm_ conflicted about this PR. 😖

www/newitems.php
alice-blue: @dfabulich can you please tell me more about what I need to do? I added a statement to alter gameRatingsSandbox0_mv to add a new column, but I'm not sure how to get the published date of the latest review into that column....it looks like maybe gameRatingsSandbox0_mv is updated from the gameRatingsSandbox0 view? So do I need to get the published date into the gameRatingsSandbox0 view? I'm kind of lost. Also, the publishing date would be the greater of either createdate or embargo date, I think, but do we only add it (where?) if the embargo date is past?

alice-blue:
Ok, I merged your PRs. (I didn't test anything.)

alice-blue:
I'm not sure how to fix the conflicts. The filetypes stuff looks familiar, but maybe from an old, unrelated PR.

dfabulich:
The idea of `incoming-schema-changes.sql` is that it's intended to take an old database dump from the IF Archive and update it to the latest database schema. `incoming-schema-changes.sql` gradually accumulates incoming changes until a new quarterly database dump arrives on the IF Archive. When I update `main` to use a newer database dump, I delete everything in `incoming-schema-changes.sql` in the `main` branch, because all of those changes had already been applied to the production database months ago.

In this PR, you just added one line to `incoming-schema-changes.sql` at the top of the file. In the `main` branch, the file changed out from under you, removing all of the lines of the file. Resolving the conflict means using the `main` blank version and re-adding just the one line you actually intended to add.

sql/incoming-schema-changes.sql
```diff
@@ -7,3 +7,101 @@ USE ifdb;
...
+
+  
+
+-- Create a materialized view to store the data from the recentgamenews view
+
+CREATE TABLE recentgamenews_mv (
+  news_id BIGINT(20) unsigned NOT NULL,
+  game_id VARCHAR(32) NOT NULL,
+  news_create_date DATETIME NOT NULL,
+  PRIMARY KEY (news_id)
```
dfabulich, line 42: Add a `KEY (game_id)` so we can do a fast `JOIN` to it.

alice-blue: Done

sql/incoming-schema-changes.sql
```diff
@@ -7,3 +7,101 @@ USE ifdb;
...
+-- superseded by later edits.
+
+CREATE VIEW `recentgamenews` AS 
+    SELECT 
+        newsid AS news_id,
+        sourceid AS game_id, 
+        created AS news_create_date
+    FROM news
+    WHERE source = "G"
+    AND newsid NOT IN (
```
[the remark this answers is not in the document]
alice-blue: Fixed.

PR #1201, opened 2024-12-06
search: Don't list negative terms in 'Tags matched'

### LINKED ISSUE #1200: Unexpected search results with partial matches of multi-word tags
If you go to IFDB and do the search  `-tag:disturbing`, some results include

> tags matched:

like this:

> Slouching Towards Bedlam
> by Star Foster and Daniel Ravipinto
> 2003
> Tags matched: disturbing themes
> (214 ratings)
 
If the term "disturbing" matches "disturbing themes," then why is this game included in search results?

On the other hand, if the tag doesn't match, then why does it say "tags matched"?

I'm not sure how search is intended to handle partial matches like this, but the "tags matched" thing kind of makes it look like it was trying to include rather than exclude the tag.

salty-horse:
Fixes #1200.

The term `(strpos($s[2], '-') !== false)` is a bit cryptic. I copied it from searchutil.php:
```php
 $negate = (strpos($s[2], '-') !== false);
```

### DIFF
--- a/www/search
+++ b/www/search
@@ -1342,6 +1342,9 @@ else if ($term || $browse)
                     $tagsMatched = array();
                     foreach ($specials as $s) {
                         if ($s[0][0] == "tags") {
+                            // Don't try to match negated terms
+                            if ($s[2] === '-')
+                                continue;
                             foreach ($tags as $t) {
                                 if (strpos(strtolower($t),
                                            strtolower($s[1])) !== false) {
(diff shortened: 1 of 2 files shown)


www/search
```diff
@@ -1342,6 +1342,9 @@ else if ($term || $browse)
                     $tagsMatched = array();
                     foreach ($specials as $s) {
                         if ($s[0][0] == "tags") {
+                            // Don't try to match negated terms
+                            if (strpos($s[2], '-') !== false)
```
dfabulich, line 1346:
```suggestion
                            if ($s[2] === '-')
```

I think `$s[2]` is necessarily either the one-character string `-` or the empty string.

https://github.com/iftechfoundation/ifdb/blob/a3a6f0b097a49e240ae30769177e9ecdc112da6a/www/searchutil.php#L345-L361

`$s[2]` is `$match[1]`, but the first matching group is `(-?)`, i.e. either a hyphen or the empty string.

dfabulich (changes requested):

PR #1211, opened 2024-12-13
Fix header search bar appearance on mobile Safari

### LINKED ISSUE #1209: Header magnifying glass doesn't vertically fit in search bar on iOS 18.2

dfabulich:
Fixes #1209

# Before

# After

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -222,9 +222,12 @@ div.topctl table tr td {
 
 .searchbar-wrapper > input {
     border: none;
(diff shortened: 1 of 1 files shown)


salty-horse:
Do you mind also testing it on Safari with `<!doctype html><html>` (set in pagetpl.php) to make sure it looks OK in standards mode?

salty-horse (approved):

dfabulich:
The header looks the same in standards mode

Issue #1213, opened 2024-12-13
Add a number badge to the Inbox / Menu header link


dfabulich:
When there are unread messages in your inbox, there's a banner on the home page, but not anywhere else.

It would be nice to add a little red number badge to the inbox, e.g. "(3)" if you have three unread messages.

Issue #917, opened 2024-10-24
PHP warnings in check-inbox.php


salty-horse:
When loading the main page, I'm getting these warnings:

```
Undefined variable: newCaughtUp in check-inbox.php on line 8
Undefined variable: srow in check-inbox.php on line 9
Trying to access array offset on value of type null in check-inbox.php on line 9
```

That piece of code used to be in `home`, and the earliest version we have in the repository doesn't mention those variables.

Any idea what that code is checking?

alice-blue:
`newCaughtUp` is used in the queryComments function (defined in commentutil.php) to track which inbox messages are considered new/unread. (There's a comment in commentutil.php saying "if there's a caught-up date, only query messages newer than that date"). I guess it checks that in order to announce "You have new messages" on the home page. I don't know what `srow` is.

PR #1215, opened 2024-12-14
Show the number of new messages next to "Inbox" in the nav bar

alice-blue:
This displays the number of new messages next to the word "Inbox" in the nav bar at the top of the page.

I experimented a little with trying to make the number stand out more by using red type, or white type on a red background, but I didn't find a combo that looked good to me (especially in dark mode), so this PR addresses only part of https://github.com/iftechfoundation/ifdb/issues/1213

inbox messages

(Originally I was testing this by logging in as different users, but I figured out that it also works if you comment on your own profile.)

### DIFF
--- a/www/pagetpl.php
+++ b/www/pagetpl.php
@@ -117,12 +117,12 @@ function pageHeader($title, $focusCtl = false, $extraOnLoad = false,
                     <li class="<?= ($pagescript === 'commentlog') ? 'page-active':''; ?>"><a id="topbar-inbox" href="/commentlog?mode=inbox">Inbox
                     <?php
                     // check the inbox so we can display the number of new messages
-                    $db = dbConnect();
-                    $uid = checkPersistentLogin();
-                    $quid = mysql_real_escape_string($uid, $db);
                     $inboxCnt = 0;
-                    if ($quid) {
-                        list($inbox, $inboxCnt) =
+                    $uid = checkPersistentLogin();
+                    if ($uid) {
+                        $db = dbConnect();
+                        $quid = mysql_real_escape_string($uid, $db);
+                        [$inbox, $inboxCnt] =
                         queryComments($db, "inbox", $quid, "limit 0, 1", $caughtUpDate, false);
                     }
                     if ($inboxCnt) { 
--- a/www/pagetpl.php
+++ b/www/pagetpl.php
@@ -117,14 +117,13 @@ function pageHeader($title, $focusCtl = false, $extraOnLoad = false,
                     <li class="<?= ($pagescript === 'commentlog') ? 'page-active':''; ?>"><a id="topbar-inbox" href="/commentlog?mode=inbox">Inbox
                     <?php
                     // check the inbox so we can display the number of new messages
-                    $inboxCnt = 0;
-                    $uid = checkPersistentLogin();
-                    if ($uid) {
-                        $db = dbConnect();
-                        $quid = mysql_real_escape_string($uid, $db);
-                        [$inbox, $inboxCnt] =
+                    $db = dbConnect();
+                    $result = mysqli_execute_query($db,
+                    "select caughtupdate from users where id=?", [$curuser]);
+                    $caughtUpDate = mysql_result($result, 0, "caughtupdate");
+                    $quid = mysql_real_escape_string($curuser, $db);
+                    [$inbox, $inboxCnt] =
                         queryComments($db, "inbox", $quid, "limit 0, 1", $caughtUpDate, false);
-                    }
                     if ($inboxCnt) { 
                         echo " (" . $inboxCnt . ")";
                     }
--- a/www/pagetpl.php
+++ b/www/pagetpl.php
@@ -3,6 +3,7 @@
 include_once "csp-nonce.php";
 include_once "util.php";
 include_once "login-persist.php";
+include_once "commentutil.php";
 
 header("Speculation-Rules: \"/speculation-rules\"");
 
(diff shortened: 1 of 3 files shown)


www/pagetpl.php
```diff
@@ -113,7 +114,22 @@ function pageHeader($title, $focusCtl = false, $extraOnLoad = false,
...
                     <li class="<?= ($pagescript === 'personal') ? 'page-active':''; ?>"><a id="topbar-personal" href="/personal">My Activity</a></li>
-                    <li class="<?= ($pagescript === 'commentlog') ? 'page-active':''; ?>"><a id="topbar-inbox" href="/commentlog?mode=inbox">Inbox</a></li>
+                    <li class="<?= ($pagescript === 'commentlog') ? 'page-active':''; ?>"><a id="topbar-inbox" href="/commentlog?mode=inbox">Inbox
+                    <?php
+                    // check the inbox so we can display the number of new messages
+                    $db = dbConnect();
+                    $uid = checkPersistentLogin();
+                    $quid = mysql_real_escape_string($uid, $db);
+                    $inboxCnt = 0;
+                    if ($quid) {
```
salty-horse, line 124: check `$uid` here, and move `mysql_real_escape_string` inside the `if`.

(Later I will change it so queryComments gets `$uid` as an argument)

www/pagetpl.php
```diff
@@ -113,7 +114,22 @@ function pageHeader($title, $focusCtl = false, $extraOnLoad = false,
...
+                        list($inbox, $inboxCnt) =
```
salty-horse, line 125: I prefer `[]` instead of `list()` in new code.

www/pagetpl.php
alice-blue: Ok, I moved that and also moved $db inside the if.

www/pagetpl.php
alice-blue: Changed this.

www/pagetpl.php
```diff
@@ -113,7 +114,22 @@ function pageHeader($title, $focusCtl = false, $extraOnLoad = false,
                     <li class="<?= ($pagescript === 'showuser') ? 'page-active':''; ?>"><a id="topbar-profile" href="/showuser">Profile</a></li>
                     <li class="<?= ($pagescript === 'editprofile') ? 'page-active':''; ?>"><a id="topbar-edit" href="/editprofile">Settings</a></li>
                     <li class="<?= ($pagescript === 'personal') ? 'page-active':''; ?>"><a id="topbar-personal" href="/personal">My Activity</a></li>
-                    <li class="<?= ($pagescript === 'commentlog') ? 'page-active':''; ?>"><a id="topbar-inbox" href="/commentlog?mode=inbox">Inbox</a></li>
+                    <li class="<?= ($pagescript === 'commentlog') ? 'page-active':''; ?>"><a id="topbar-inbox" href="/commentlog?mode=inbox">Inbox
+                    <?php
+                    // check the inbox so we can display the number of new messages
+                    $inboxCnt = 0;
+                    $uid = checkPersistentLogin();
```
dfabulich, line 121: We already called `checkPersistentLogin()` at line 82. We probably don't want to do it again here.

www/pagetpl.php
```diff
@@ -113,7 +114,22 @@ function pageHeader($title, $focusCtl = false, $extraOnLoad = false,
...
+                    if ($uid) {
```
dfabulich, line 122: I think we actually don't need an `if` statement here, because we checked if we're logged in on line 113.

www/pagetpl.php
```diff
@@ -113,7 +114,22 @@ function pageHeader($title, $focusCtl = false, $extraOnLoad = false,
...
+                        $db = dbConnect();
+                        $quid = mysql_real_escape_string($uid, $db);
+                        [$inbox, $inboxCnt] =
+                        queryComments($db, "inbox", $quid, "limit 0, 1", $caughtUpDate, false);
```
dfabulich, line 126: This line isn't working right. I see this warning in the Docker console:

```
PHP Notice:  Undefined variable: caughtUpDate in /var/www/html/pagetpl.php on line 125
```

And, sure enough, the number flagged in the inbox is the count of all comments for all time; even if you go to the inbox and click "I'm caught up," the Inbox header number doesn't decrease or go away.

`home` sets `$caughtUpDate` on line 27. I think we'd need to do something similar here in `pagetpl.php`.

dfabulich (changes requested):
We've got to fix the `caughtUpDate` before we should merge this.

dfabulich:
It would also be nice if we could skip calling `queryComments` in `check-inbox.php`, but that seems kinda tricky. (Maybe use a `global`?)

salty-horse:
See also #917

dfabulich:
I filed https://github.com/alice-blue/ifdb/pull/7 to show the Inbox count in a red number.

salty-horse:
If the number gets too big, it would look bad in a circle. Maybe it should cap at "99+".

www/pagetpl.php
alice-blue: Ok, I removed this.

www/pagetpl.php
alice-blue: Ok, $caughtUpDate is now set.

www/pagetpl.php
alice-blue: I changed references to $uid to say $curuser instead

alice-blue:
> It would also be nice if we could skip calling `queryComments` in `check-inbox.php`, but that seems kinda tricky. (Maybe use a `global`?)

@dfabulich I wasn't really sure how to do this, but tried anyway. I deleted the call to queryComments from check-inbox, and also added globals (I think?) in commentutil.php.

dfabulich (changes requested):
Minor conflict with #1217

alice-blue:
> Minor conflict with #1217

I think I unconflicted it.

PR #1216, opened 2024-12-14
 Fix error handling in reviewvote, only show buttons when logged in @salty-horse

### LINKED ISSUE #1192: Review vote UI pretends to let you vote when you're logged out
Log out and navigate to any review.

There are visible links, "Was this review helpful to you?   Yes   No". Click "Yes"

Expected: You should be required to login in order to vote on a review.
Actual: The UI updates to say "(You voted Yes)"

It doesn't display the red bold "Vote Recorded" UI message (not that the user would know that it's missing). If you check Chrome Dev Tools, the vote is rejected with a 401 HTTP status code.

salty-horse:
Fixes #1192.

* jsonSend didn't handle some errors correctly. It didn't show the alert message when there was an HTTP error, and it always called the callback. IIRC the intent is to only call the callback when there are no HTTP errors. (The callback can't tell if it was successful or not). Now it will still call it if there's an `error` field in the JSON body.
* When a user is logged out, they will not be presented with the review vote controls.

### DIFF
--- a/www/ifdbutil.js
+++ b/www/ifdbutil.js
@@ -238,8 +238,9 @@ async function jsonSend(url, statusSpanID, cbFunc, content, silentMode)
 
     let jsonResponse = null;
     let msgspan = null;
+    let response;
     try {
-        const response = await fetch(url, options);
+        response = await fetch(url, options);
         jsonResponse = await response.json();
         msgspan = (statusSpanID
                 ? document.getElementById(statusSpanID)
@@ -265,7 +266,7 @@ async function jsonSend(url, statusSpanID, cbFunc, content, silentMode)
                    + "(" + response.status + ") "
                    + "Please try again later.");
     }
-    if (cbFunc) {
+    if (cbFunc && response?.ok) {
         cbFunc(jsonResponse);
     }
 }
--- a/www/reviews.php
+++ b/www/reviews.php
@@ -383,7 +383,9 @@ function showReview($db, $gameid, $rec, $specialNames, $optionFlags = 0)
 
     } else {
         // not special - show the headline and author
-        echo " <b>$summary</b><span class=details>, $publicationdate</span><br>"
+        echo " <b>$summary</b><span class=details>, "
+            . "<a class=\"silent\" href=\"viewgame?id=$gameid&review=$reviewid\" "
+            . "title=\"Direct link to this review\">$publicationdate</a></span><br>"
             . "<div class=smallhead><div class=details>"
             .   "by <a href=\"showuser?id=$userid\">$username</a>"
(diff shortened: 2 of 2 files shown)


dfabulich (changes requested):
In #1192, you wrote:

> The voting UI should be hidden when you're not logged in. (IMO - I prefer that to trying to do the action and being redirected to a login page.)

I think I don't agree with that approach. When logged out, there should be _some_ invitation to login to use features, rather than hiding the features completely.

For example, "Have you played this game?" says, "You can rate this game, record that you've played it, or put it on your wish list after you log in." We can put a message like that here, e.g. "You can log in to rate this review, mute this user, or add a comment."

That doesn't always fit in the UI, and so a good enough second-choice option is to leave the links in place but require login, as we do for "Add a Comment".

Regardless, this PR is currently doing something kinda weird, hiding _just_ the review vote options, but showing "More Options" (all of which require login, except for "Direct Link") and Add a Comment (which redirects to login).

(And, why would it be called "More Options" when the review vote options are hidden…?)

salty-horse:
Mostly agreed.

I don't think every review should have a "call to action" to login under it, when there can be several reviews on a page.

I missed that detail about "More Options" only having login-requiring actions. I originally put the "Direct Link" in the "More" menu and I regret it. Got any suggestions on where to place it? Usually permalinks are placed on the timestamps but that's not ideal.

dfabulich:
> I don't think every review should have a "call to action" to login under it, when there can be several reviews on a page.

I think it's OK (good) to show the call-to-action message next to every review, (as long as the message itself is brief) because I think it's unlikely that we'd show the call to action multiple times on the same screen. The median review is ~600-700 words, a few dozen lines long. So I think the user will likely see it only once (or at most twice) in any given screenful.

If we're hiding "More Options" when logged out, then I think the thing to do is to add the permalink to the timestamp, but I think it's OK to leave it in "More Options", too. It's redundant, but it's not occupying extra screen space, and the redundancy would make the permalink a little bit easier to discover.

salty-horse:
Rebased over main to include the plonk->mute changes.

1. Added a "Log in to do stuff" message with your suggested phrasing. The redirect after login leads to the specific review, rather than the game page.
2. The "more options" box only appears when you're logged in.
3. The timestamps in reviews is now a permalink with the "silent" style, so it's not blue and only has an underscore on hover. That permalink is only on reviews with text - should they be included in rating-only reviews? The permalink is always there, even when viewing a specific review, because I expect permalinks to always be there in case I want to copy them.

dfabulich:
> That permalink is only on reviews with text - should they be included in rating-only reviews?

I guess it would be nice, but I wouldn't hold up this PR over it.

PR #1220, opened 2024-12-16
Show Inbox number on mobile

dfabulich:
# Menu closed

image

# Menu open

Note: We're showing the number twice. Is that bad? I guess I prefer it…?

image

# Dark mode

image

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -2178,7 +2178,7 @@ span.sfl-save-autodesc {
     background: #b0d8e8;
 }
 
-#topbar-inbox .notification {
+.inbox-notification {
     min-width: 1.6em;
     height: 1.6em;
     border-radius: 0.8em;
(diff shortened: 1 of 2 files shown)


dfabulich:
@alice-blue

alice-blue:
Having the number next to "inbox" in the drop-down menu makes sense. I'm not sure how understandable it will be to also have it show up next to menu. (There are 17 menu items? 17 _new_ menu items?) But I guess people will figure it out if they click menu, and then see the matching number next to inbox.

dfabulich:
Another idea I toyed with was to hide the number in the "Menu" button when you expand the menu, but that made the Menu scoot around in a way that didn't feel good. This was the best option I could come up with, but I still don't feel totally satisfied with it.

alice-blue:
I don't think it's a big deal to have the number shown twice. I'd be a little more concerned about confusion than redundance. The only other thing I can think of is to put some kind of icon there instead of a circle (maybe an envelope icon? A red one, even?) and have alt text for it. But this also seems like the kind of thing we can improve later if we think of a better idea.

PR #1222, opened 2024-12-17
Add URL for Play Online button to the viewgame API

alice-blue:
Add "playOnlineMainURL" to the viewgame API. This URL should be identical to the URL that's used in the Play Online button.

The new code is in lines 820-839. The other section is--if I've done it right--existing code moved earlier in the page.

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -572,13 +572,14 @@ foreach ($links as $link)
         'tads3',
         'hugo']))
     {
+        $parchment = "https://iplayif.com/?story=";
         if (!$link['compression']) {
             $foundPC = true;
-            $pcUrl = urlencode(urlToMirror($link['url']));
+            $pcUrl = $parchment . urlencode(urlToMirror($link['url']));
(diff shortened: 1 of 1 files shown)


www/viewgame
```diff
@@ -506,6 +506,110 @@ if ($mediantime > 60) {
...
+    {
+        if (!$link['compression']) {
+            $foundPC = true;
+            $pcUrl = urlencode(urlToMirror($link['url']));
+            break;
+        } else if ($unboxUrl) {
+            $foundPC = true;
+            $pcUrl = urlencode($unboxUrl);
+            break;
+        }
```
salty-horse, line 583: `$pcUrl` should be modified to have the `https://iplayif.com/?story=` prefix here, instead of what is now two places.

salty-horse: Basically, it should work like the adrift code.

alice-blue: Ok, I changed this.

PR #1224, opened 2024-12-18
Lots of play online urls

alice-blue:
This PR takes some steps towards including "play online" urls for every relevant link for a given game in the API. Unfortunately, it's buggy. The API is showing the same play online url for every link for a given game.

PR #1226, opened 2024-12-18
Add small Play Online links to every online-playable link

### LINKED ISSUE #1121: Add "play online" links to all story files with supported web interpreters
The External Links section has several usability problems regarding Play Online.

1. Only the first link marked "This is a playable game file or it contains a playable game" gets a Play Online link. This "Play Online" link is the only way story files gets wrapped up with web interpreters for easy use. Visitors who don't want to mess with running games locally, or don't have sufficient know-how to point iplayif at the provided files, have no way of trying those additional files.
2. When editing an external link, the bottom of the dialog box says:
    > According to the settings above, this link will include a "Play On-line" button.

   This is not always true. If the game the second such-marked file, it will not get a button. (BTW, the hyphen should be removed).

I think that all playable links that have web interpreters (Parchment-capable or ADRIFT) should get a "Play online" link next to them. It doesn't need to be as prominent as the regular "Play online" button.

dfabulich:
Fixes #1121

This is easy now that we've merged @alice-blue's PR #1224.

### DIFF
--- a/www/editgame
+++ b/www/editgame
@@ -797,10 +797,10 @@ function updateFormatUI(byUser)
         priNA.style.display = "none";
     }
 
-    var noButton = "According to the settings above, this link will <b>not</b> include a \"Play On-line\" button, because ";
+    var noButton = "According to the settings above, this link will <b>not</b> include a \"Play Online\" link, because ";
 
     if (/Play Online/i.test(document.getElementById("linktitle").value)) {
-        playOnlineExplanation.innerHTML = "According to the settings above, this link will include a \"Play On-line\" button.";
+        playOnlineExplanation.innerHTML = "According to the settings above, this link will include a \"Play Online\" link.";
     } else if (!selIsg.checked) {
         playOnlineExplanation.innerHTML = noButton + 'the "playable game file" option is not checked.';
     } else if (!selFmtG.value) {
@@ -815,7 +815,7 @@ function updateFormatUI(byUser)
     } else if (selCmp.value === '17' && !priFld.value) {
         playOnlineExplanation.innerHTML = noButton + 'the "Main File" is blank.';
     } else {
-        playOnlineExplanation.innerHTML = "According to the settings above, this link will include a \"Play On-line\" button.";
+        playOnlineExplanation.innerHTML = "According to the settings above, this link will include a \"Play Online\" link.";
     }
 
     if (byUser && fmt && fmt.externid == "tads3web")
@@ -1447,11 +1447,11 @@ foreach ($linkfmts as $lf) {
                </tr>
                <tr valign=top>
                   <td align=right>
-                     <b><nobr>"Play On-line" button:</nobr></b>
+                     <b><nobr>"Play Online" link:</nobr></b>
                   </td>
                   <td>
                      <span id="playOnlineExplanation">
-                        According to the settings above, this link will <b>not</b> include a "Play On-line" button, because the "playable game file" option is not checked.
+                        According to the settings above, this link will <b>not</b> include a "Play Online" link, because the "playable game file" option is not checked.
                      </span>
                   </td>
                </tr>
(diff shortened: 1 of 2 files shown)


salty-horse:
I was thinking maybe those links should be styled like the "button", and floated to the right. They can be smaller than the main button.

Open issues from my report:
1. The "Edit External Link" dialog says "According to the settings above, this link will include a "Play On-line" button." -- now it's a button or link.
2. Should there be a way to disable the main "play online" button, for cases where there's no true "main" version of the game?

alice-blue:
little play online links

Now that people can play the same version of a game either online or offline, the sequence of story file, the play online link, and the "Requires a Z-Code interpreter" note seems potentially confusing. I wonder if there's a way the downloadable story file could be more clearly paired with the "interpreter is required" note, somehow, and the play online link could be presented as an alternative to downloading an interpreter.
Would it make sense to say 
```
Advent.z5
350 points
Requires a Z-Code interpreter. Visit IFWiki for download links, or play online.
```
or maybe


```
Advent.z5
350 points
* To play offline, you'll need a Z-Code interpreter. Visit IFWiki for download links.
* Play online.
```

dfabulich:
> 1. The "Edit External Link" dialog says "According to the settings above, this link will include a "Play On-line" button." -- now it's a button or link.

Fixed. Now it's called the "Play Online link." (Even if it's a button, the button is still basically just a link, so it should be accurate regardless.)

> I was thinking maybe those links should be styled like the "button", and floated to the right. They can be smaller than the main button.

I experimented with this a little, but I couldn't make a version that I liked, especially when the first link is playable online.

image

> 2. Should there be a way to disable the main "play online" button, for cases where there's no true "main" version of the game?

Filed as https://github.com/iftechfoundation/ifdb/issues/1230

> Advent.z5
> 350 points
> * To play offline, you'll need a Z-Code interpreter. Visit IFWiki for download links.
> * Play online.

I don't think I like this version, because it might make the "Play online" link appear to apply to the subsequent link.

> Advent.z5
> 350 points
> Requires a Z-Code interpreter. Visit IFWiki for download links, or play online.

This one is interesting, but I think it would be more work than I'm prepared to do right now.

Currently, the "Requires a Z-Code interpreter …" line is stored in the database, in the `filetypes` table.

```
MariaDB [ifdb]> select id, `desc` from filetypes where externid = 'zcode';
+----+--------------------------------------------------------------------------------------------------------------------------+
| id | desc                                                                                                                     |
+----+--------------------------------------------------------------------------------------------------------------------------+
|  6 | Requires a Z-Code interpreter. Visit <a href="https://www.ifwiki.org/Z-code_interpreters">IFWiki</a> for download links. |
+----+--------------------------------------------------------------------------------------------------------------------------+
```

To make it look as you suggest, I'd have to edit all of the online-playable descriptions to something like `Requires a Z-Code interpreter. Visit IFWiki for download links, or #PLAY_ONLINE#` and then add code to replace `#PLAY_ONLINE#` with `<a href="$playOnlineUrl">play online</a>`.

I'm open to that in a separate PR, but I think this is at least good enough for now.

alice-blue:
Here's another option: https://github.com/iftechfoundation/ifdb/pull/1231

dfabulich:
I posted some suggested fixes to #1231, but I honestly kinda prefer my version here in #1226 better. @salty-horse do you have a preference?

salty-horse:
@dfabulich I like having the "Play online" buttons next to each link, as depicted in your screenshot. I think the top one next to "External links" title can be removed. It's sometimes a mystery which of the links it applies to, and having it next to the link makes it obvious which version of the game it will lead to.

However, I know the top button moves into the main game info box when in mobile mode, and maybe that should remain. And in terms of accessibility, this requires more "scanning" of the page to reach the button correct button.

dfabulich:
I think I'm convinced to go with #1233

PR #1231, opened 2024-12-19
Another option for the play online mini links

alice-blue:
Similar to what I suggested, but this version doesn't require editing any existing messages, I don't think. It adds a separate sentence saying "You can also play online."

### DIFF
--- a/www/viewgame
+++ b/www/viewgame
@@ -1318,14 +1318,15 @@ if (count($links) == 0 && $dlnotes == "") {
             echo "<span class=dlnotes>$linkdesc<br></span>";
         }
 
-        // Don't show "Play Online" mini link for HTML links; that's what the link itself does
-        // ... unless the HTML is zipped, in which case we *do* need a Play Online link
         $playOnlineMiniLink = "";
-        if ($playOnlineURL && ($onlineInterpreterType !== "HTML" || $compPri)) {
-            if ($typedesc || $zipnote) {
-                $playOnlineMiniLink = " You can also <a href='$playOnlineURL'>play online</a>.";
+        if ($playOnlineURL) {
+            if ($onlineInterpreterType === "HTML" && !$compPri) {
+                // Don't show "Play Online" mini link for HTML links; that's what the link itself does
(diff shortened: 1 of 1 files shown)


www/viewgame
```diff
@@ -1312,20 +1312,25 @@ if (count($links) == 0 && $dlnotes == "") {
...
+        // Don't show "Play Online" mini link for HTML links; that's what the link itself does
+        // ... unless the HTML is zipped, in which case we *do* need a Play Online link
+        $playOnlineMiniLink = "";
+        if ($playOnlineURL && ($onlineInterpreterType !== "HTML" || $compPri)) {
+            if ($typedesc || $zipnote) {
+                $playOnlineMiniLink = " You can also <a href='$playOnlineURL'>play online</a>.";
+            } else {
+                $playOnlineMiniLink = "<a href='$playOnlineURL'>Play Online</a>";
+            }
+        }
```
dfabulich, line 1330:
```suggestion
        $playOnlineMiniLink = "";
        if ($playOnlineURL) {
            if ($onlineInterpreterType === "HTML" && !$compPri) {
                // Don't show "Play Online" mini link for HTML links; that's what the link itself does
                // ... unless the HTML is zipped, in which case we *do* need a Play Online link
            } else if ($zipnote || !$typedesc || $onlineInterpreterType === "HTML") {
                $playOnlineMiniLink = " <a href='$playOnlineURL'>Play Online</a>";
            } else {
                $playOnlineMiniLink = " You can also <a href='$playOnlineURL'>play online</a>.";
            }
        }
```

dfabulich (changes requested):
Suggested replacement. This works for tricky cases I tried.

dfabulich:
https://github.com/iftechfoundation/ifdb/pull/1226#issuecomment-2552702002

> I honestly kinda prefer my version here in https://github.com/iftechfoundation/ifdb/pull/1226 better. @salty-horse do you have a preference?

salty-horse:
@dfabulich I prefer the button-next-to-title in the #1226 screenshot, and as a second choice the link-just-below-title.

dfabulich:
Let's go with #1233 instead

PR #1234, opened 2024-12-23
Add view transitions to coverart on game details pages

dfabulich:
The view transition API is brand new in 2024. It released just a couple weeks ago in Safari 18.2 and in June in Chrome 126. It's not available in Firefox yet, but that's fine, because it's a progressive enhancement.

(I'm not even sure this is a good idea to implement, but I wanted to play around with this API, so, here we are.)

https://github.com/user-attachments/assets/1b24e445-7524-4323-9def-133d7e60cfa1

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -60,6 +60,16 @@ html {
     scroll-behavior: smooth;
 }
 
+@view-transition {
+    navigation: auto;
+}
+
+@media (prefers-reduced-motion: reduce) {
+    @view-transition {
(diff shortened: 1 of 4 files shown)


salty-horse:
Didn't know this existed. Nice!

Feels a bit over-the-top for this use-case - do you think it improves the browsing experience?

salty-horse (approved):

dfabulich:
I've thought it over, and I think this isn't a good use of animations.

PR #1239, opened 2025-01-03
Games with genres/tags in common -- experimental feature

alice-blue:
This is probably not ready to merge. But aside from some cleanup, I'm not sure what to focus on.


alice-blue:
I cleaned it up some more.

PR #1249, opened 2025-01-16
ifarchive-scrubber: Scrub reviews from users who are banned and sandboxed

### LINKED ISSUE #1248: IF Archive scrubber doesn't delete reviews from banned, sandboxed users
In the dev environment, query for reviews from users that don't exist in the system.

```
MariaDB [ifdb]> select count(reviews.id)
    -> from reviews
    -> left outer join users on userid = users.id
    -> where
    ->     review is not null
    ->     and userid is not null
    ->     and userid not in ('', '$system')
    ->     and users.id is null;
+-------------------+
| count(reviews.id) |
+-------------------+
|               106 |
+-------------------+
```

A number of these reviews are from users who are banned and sandboxed.

dfabulich:
Fixes #1248

### DIFF
--- a/sql/ifdb-to-ifarchive-scrubber.sql
+++ b/sql/ifdb-to-ifarchive-scrubber.sql
@@ -18,9 +18,6 @@ drop table reviewflags;
 drop table userfilters;
 drop table stylepics;
 
-delete from users
-  where acctstatus != 'A';
-
 update users
   set picture = concat('https://ifdb.org/showuser?pic&id=', id)
   where picture is not null;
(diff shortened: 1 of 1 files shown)


sql/ifdb-to-ifarchive-scrubber.sql
```diff
@@ -154,3 +151,7 @@ drop table gametimes_mv;
 drop trigger playertime_insert;
 drop trigger playertime_update;
 drop trigger playertime_delete;
+
+-- delete users with acctstatus D (pending activation) R (pending review) B (banned)
+delete from users where acctstatus != 'A';
```
salty-horse, line 156: If  you mean D/R/B and expect other statuses in the future, then maybe it's better to use `IN ('D', 'R', 'B')`

salty-horse (approved):

PR #1256, opened 2025-01-18
Ensure all games appear in `gameRatingsSandbox0` view

### LINKED ISSUE #1255: Some games aren't being counted when browsing all games
On the home page, scroll down to the "Database Stats." As I write this, it says we have 14084 game listings. Click the link.

Actual: In the upper-right corner of the browse page, it will show a smaller number of games. As I write this, it says "1–100 of 13234"

Expected: The Browse page should match the total number of games.

dfabulich:
We were filtering out special reviews, embargoed reviews, and sandboxed reviews with the `where` clause, but this meant that games that had reviews but _only_ special reviews were excluded from the view completely.

Now, we've moved the filters into the criteria for the `left outer join`s. If there happen to be no rows that match those criteria, that's fine, it's an outer join, so we'll still have at least one row per game in the `grouped` query results.

For example, zz69axowzyl0htmh "Murder Mansion" had just one special review. Previously, the `grouped` query returned 0 rows for this game, now it returns one row with `count: 0` `rating: NULL`.

Fixes #1255

### DIFF
--- a/sql/unscrub-ifarchive.sql
+++ b/sql/unscrub-ifarchive.sql
@@ -429,15 +429,19 @@ from (
               `ifdb`.`reviews`.`review` is not null AS `hasReview`
             from (
                 `ifdb`.`games`
-                left outer join `ifdb`.`reviews` on (`ifdb`.`games`.`id` = `ifdb`.`reviews`.`gameid`)
-                left outer join `ifdb`.`users` on(`ifdb`.`reviews`.`userid` = `ifdb`.`users`.`id`)
+                left outer join `ifdb`.`reviews` on (
+                  `ifdb`.`games`.`id` = `ifdb`.`reviews`.`gameid`
+                  and ifnull(
+                    current_timestamp() > `ifdb`.`reviews`.`embargodate`,
+                    1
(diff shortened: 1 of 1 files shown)


sql/unscrub-ifarchive.sql
```diff
@@ -429,15 +429,19 @@ from (
...
-                1
+                left outer join `ifdb`.`reviews` on (
+                  `ifdb`.`games`.`id` = `ifdb`.`reviews`.`gameid`
+                  and ifnull(
+                    current_timestamp() > `ifdb`.`reviews`.`embargodate`,
+                    1
+                  )
+                  and `ifdb`.`reviews`.`special` is null
+                )
+                left outer join `ifdb`.`users` on(
```
salty-horse, line 440: add space between `on(`

sql/unscrub-ifarchive.sql
```diff
@@ -596,15 +600,18 @@ from (
...
```
salty-horse, line 611: add space between `on(`

salty-horse (approved):

PR #1259, opened 2025-01-18
Suppress buried game warning on first Browse page

dfabulich:
Fixing #1256 seems to have made some buried games appear on the first page of Browse results. Let's just suppress the warning on the first page of browse results.

### DIFF
--- a/www/search
+++ b/www/search
@@ -1267,7 +1267,7 @@ else if ($term || $browse)
         echo "</div>";
 
         $showFlagged = isset($_GET['showFlagged']) && $_GET['showFlagged'];
-        if (!$showFlagged && $searchType == "game") {
+        if (!$showFlagged && $searchType == "game" && (!$browse || $pg != 1)) {
             for ($i = 0 ; $i < count($rows) ; $i++) {
                 $row = $rows[$i];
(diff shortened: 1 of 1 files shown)


salty-horse:
I don't fully understand what this means. Can you provide a screenshot of this warning, and why these games should not be visible?

Also, IMO it's cryptic enough to merit a comment in the code.

dfabulich:
Here's a screenshot below of the banner. ("Some results were hidden. <ins>See all results</ins>")

Here's an example buried game, Necromancer. <https://ifdb.org/viewgame?id=b36l51ogn147x0iv> Necromancer was buried due to coordinated vote manipulation. You can still see the banner if you search for Necromancer. <https://ifdb.org/search?searchbar=necromancer>

I'm hiding the banner on the first page of the Browse Games page, because the banner would be too prominent in that case.

image

dfabulich:
It turns out that I didn't actually need to do this, because buried games were only appearing in the top 100 because of #1263. Now that that's fixed, no buried games naturally appear on the first page of Browse Games.

But, I still want to suppress the banner on the first page of Browse Games, just in case this crops up again.

PR #1261, opened 2025-01-19
Add $count_all_possible_rows parameter to doSearch

alice-blue:
This starts with the [filter games version 3 branch](https://github.com/iftechfoundation/ifdb/pull/1199), and then adds a "$count_all_possible_rows" parameter to the doSearch function. (Currently, only the last 5 commits are new. All the others are from the filter games v 3 branch.)

### DIFF
--- a/sql/incoming-schema-changes.sql
+++ b/sql/incoming-schema-changes.sql
@@ -11,3 +11,14 @@ alter table reviewvotes
 -- Add column for game search filter to the users table
 
 ALTER TABLE `users` ADD COLUMN `game_filter` VARCHAR(150) DEFAULT '';
+
+
+
+-- Add column for the publication date of the most recent review to the 
+-- game ratings sandbox materialized view
+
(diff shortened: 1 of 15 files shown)


alice-blue:
When I apply a filter, the reviews on the front page disappear, and I get a PHP fatal error message in the terminal, complaining about line 1218 in searchutil.php, which is

`            $rowcnt = length($rows);`

> PHP Fatal error:  Uncaught Error: Call to undefined function length() in /var/www/html/searchutil.php:1218\nStack trace:\n#0 /var/www/html/newitems.php(199): doSearch()\n#1 /var/www/html/components/reviews.php(4): getNewItems()\n#2 /var/www/html/home(157): include('/var/www/html/c...')\n#3 {main}\n  thrown in /var/www/html/searchutil.php on line 1218

I didn't see this error in the filter games 3 branch, and I'm not sure what led to it exactly, but if I change it to say 

`$rowcnt = count($rows);`

then the reviews come back.

dfabulich:
I rebased the new commits into PR #1265.

Issue #1268, opened 2025-01-24
On home page, the IFDB Recommends "More recommendations" link doesn't account for whether you're logged in


alice-blue:
It leads to a search for "played:no willplay:no wontplay:no reviewed:no rated:no" even if you're logged out. Should it?

dfabulich:
In #1269, we discussed this, and we decided it's OK for the search terms to stay (they just don't do anything when you're logged out).

PR #1269, opened 2025-01-24
When clicking on "More recommendations" while logged out, don't use…

alice-blue:
search terms that require being logged in.

Also, move "More recommendations" to the end to be consistent with the other components on the home page. (In the other section headlines, the headline space is used for the headline itself and possibly a link for adding new things to the database. "More recommendations" isn't that type of link.)

### DIFF
--- a/www/components/ifdb-recommends.php
+++ b/www/components/ifdb-recommends.php
@@ -38,9 +38,8 @@ if (count($recs) >= 2) {
     usort($recs, "sortBySortorder");
 
     // start the section
-    echo "<div class='headline' id='ifdb-recommends'><h1 class='unset'>IFDB Recommends</h1>"
-      . "<span class='headlineRss'><a href='/search?searchbar=played%3Ano+willplay%3Ano+wontplay%3Ano+reviewed%3Ano+rated%3Ano'>More recommendations</a></span>"
-      . "</div><div>";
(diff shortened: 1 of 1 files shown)


dfabulich:
I'm happy to move the link, but in #1268, you ask, "should it?" and I'm not really sure.

It seems "simpler" to just always link to a search for `played:no willplay:no wontplay:no reviewed:no rated:no`, and the link _does_ work for logged out users; it just doesn't accomplish much, because all of those search filters silently do nothing when you're logged out.

@salty-horse Do you have an opinion about whether we should link to `played:no willplay:no wontplay:no reviewed:no rated:no` when you're logged out?

alice-blue:
Possibly I was thinking that the extra prefixes could slow down the query unnecessarily, but I guess that's not the case?

The way the link works now, if someone uses it while logged out, it might be confusing when those willplay/wontplay/played search terms show up on the screen but don't seem to do anything. On the other hand, if those search terms _don't_ show up, then people might never realize that recommendations are supposed to take willplay/wontplay/played into account, and they won't think to log in to get better recommendations.

So maybe it's better to just show the search terms regardless.

salty-horse (commented):
It's technically still "more recommendations", no? I'm OK with keeping the current search terms.

dfabulich:
Closing this, since we've merged #1276 and we've closed #1268 as unplanned.

PR #1278, opened 2025-02-07
util.php: fix convertTimeToText function

alice-blue:
The convertTimeToTextfunction was including an extra space between the number and the word "minutes" (e.g. "5  minutes"), but it wasn't visible unless you looked at the page source. (Do extra spaces get automatically ignored when displayed as HTML?)

To see the difference, go to for example /viewgame?id=2wbz3y8neawpd64z and view the page source.

### DIFF
--- a/www/util.php
+++ b/www/util.php
@@ -2836,9 +2836,11 @@ function collapsedAuthors($authors) {
     return $str;
 }
 
-
+// ----------------------------------------------------------------------------
+//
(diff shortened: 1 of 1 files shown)


salty-horse (approved):

dfabulich:
> (Do extra spaces get automatically ignored when displayed as HTML?)

Yes, they do.

PR #1282, opened 2025-02-10
Prevent IFDB Recommends from recommending buried games

### LINKED ISSUE #1281: IFDB Recommends will recommend buried games
In a dev environment, run this SQL query:

```
insert into unwishlists (gameid, userid)
select games.id as gameid, '0000000000000000' as userid
from games where flags != 1;
```

When we bury a game we set `games.flags` to 1, `FLAG_SHOULD_HIDE` in `util.php`. This query marks all games in the DB as "Not Interested" except for the buried games. (I encountered this organically as I was impersonating a user.)

Login as ifdbadmin and view the "IFDB Recommends" section of the home page.

Actual: IFDB Recommends will recommend buried games.
Expected: IFDB Recommends should never recommend buried games.

dfabulich:
Fixes #1281

### DIFF
--- a/www/components/ifdb-recommends.php
+++ b/www/components/ifdb-recommends.php
@@ -29,6 +29,10 @@ list($recs, $rowcnt, $sortList, $errMsg, $summaryDesc, $badges,
     $specials, $specialsUsed, $orderBy) =
     doSearch($db, $term, $searchType, $sortby, $limit, $browse, $count_all_possible_rows);
 
+$recs = array_values(array_filter($recs, function($r) {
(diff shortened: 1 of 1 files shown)


dfabulich:
I think a "better" way to handle this would be to add a `buried:no` search filter, but that's more work, and I wouldn't want to support `buried:yes` searches, so we'd have to add some hackery to prevent that. Doesn't seem worth the trouble in this case.

alice-blue (approved):
I tested briefly and it seems to work.

salty-horse (commented):
Looks fine, but can this possibly reduce the list to having 0 items?

Issue #834, opened 2021-02-09
Mod tools should include 'recent activity' with all activity listed


brirush84:
This should exclude private messages but include ratings, poll votes, page edits, tags, and 'helpful' review votes

PR #1284, opened 2025-02-15
adminops: Add "recent activity by all users" tool

alice-blue:
Fixes https://github.com/iftechfoundation/ifdb/issues/834

This PR adds a tool to the admin panel that shows recent activity for all users in five categories: ratings/reviews, poll votes, page edits, tags, and 'helpful' review votes.

When testing, please change the $days variable (line 95) to a number that corresponds to how old the most recent data is. E.g. if the newest available data is 80 days old, and you want to show three days of data, set it to 83.

Known issues:

1. Before merging, $days should be changed to 3, since that's how many days of recent activity the mods requested.
2. There's supposed to be a "no [type of activity] was found" message that displays when there is no activity of that type to show, and it doesn't always display, even if you reduce $days to the point that there's no activity to show.  I don't know why.
3. The review helpfulness votes section doesn't work, and I don't know how to test it (that voter data is not public).
4. I haven't tried to tackle or test anonymous poll votes.

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -75,6 +75,329 @@ if (isset($_REQUEST['adminemail'])) {
     exit();
 }
 
+
+//------------------ RECENT ACTIVITY BY ALL USERS --------------
+
+// For long author and title strings, cut off the 
+// string after a set number of characters and 
+// add "..." if any characters were removed.
+function shortenString($long_string) {
+    $short_string = substr($long_string,0,50);
+    if ($short_string != $long_string) {
+        $short_string .= "...";
+    }
+    return $short_string;
+}
+
+
+if (isset($_REQUEST['recentactivity'])) {
+    pageHeader("Recent activity by all users");
+    $days = 79; // how many days of recent activity to display
+    echo '<h1 id="recent-activity">Recent activity by all users</h1>';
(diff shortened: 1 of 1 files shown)


alice-blue:
I experimented with anonymous poll votes, creating a test poll with anonymous voting to test it.

The recent activity code is now set up so that if the poll is anonymous, the user id is shown instead of the voter name (but mods can click through to see the name).

@dfabulich I was guessing that when the flags column in the vote table says `1`, that means the poll is anonymous, but you can tell me if that's wrong.

I see this in util.php

`define("POLL_FLAG_ANONYMOUS", 0x0001);`

and I see this in poll:

`$anonymous = $flags & POLL_FLAG_ANONYMOUS;`

dfabulich (approved):

alice-blue:
> The limit on the reviewvotes query (currently in line 211)
>         ORDER BY reviewvotes.createdate DESC LIMIT 100', [$recent_days]);
> exists only because of https://github.com/iftechfoundation/ifdb/issues/1286. The limit should probably be removed after the createdate column in the reviewvotes table gets sorted out.

@dfabulich Do you want me to do this now?

dfabulich:
Let's do it in a separate PR, I think, which could be a stacked PR, or just wait until we've deployed these changes.

PR #1294, opened 2025-04-11
Prevent logo from overlapping navbar

### LINKED ISSUE #1247: Home page: IFTF logo overlaps nav bar when scrolling
This is a minor thing, but when I start to scroll down from the top of the home page, and the IFTF logo gets near the top edge of the screen, the logo goes in front of the "Games Reviews Competitions" etc. nav bar.

IFTF logo overlap

Roachbones:
Fixes #1247. The logo's `filter` property was causing it to render over the navbar. So I gave the navbar a z-index of 1. But the mobile menu already had a z-index of 1, so I bumped that up to 2.

### DIFF
--- a/www/ifdb.css
+++ b/www/ifdb.css
@@ -1367,7 +1367,7 @@ span.headlineRss {
         top: calc(55px + 2.5em);
         right: 0;
         background-color: #e8e8e8;
-        z-index: 1;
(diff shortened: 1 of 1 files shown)


dfabulich:
Thanks for this fix!

PR #1302, opened 2025-07-07
Lower caps

brirush84:
Right now, typing "System:adventuron" into the search bar provides results but just searches for the words 'system' and 'adventuron'. Typing "system:adventuron" correctly returns games in the adventuron system.

This update is intended to reduce the incoming data to lowercase. I've tested it in codespace and it seems to work. 

I was worried it might mess up non-standard characters, but it seems like we have trouble with non-standard characters anyway. For instance, searching for "年" doesn't bring up 年獸文字冒險遊戲 | The Beast, Nian: A Chinese Text Adventure.

This is my first pull request in a while (I took a gihub class for certification and want to see if it helped make this all make more sense), so feel free to let me know if this is the wrong way to do things.

### DIFF
--- a/www/util.php
+++ b/www/util.php
@@ -239,7 +239,7 @@ function get_req_data($id)
     $val = (isset($_POST[$id]) && $_POST[$id]) ? $_POST[$id] :
            (isset($_REQUEST[$id]) ? $_REQUEST[$id] : "");
     //Convert to lowercase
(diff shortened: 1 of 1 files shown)


brirush84:
Realized that this also lowercased all input into game fields. Will continue to investigate.

PR #1312, opened 2025-08-14
Re-organize the links in the admin control panel

alice-blue:
When working on another PR, I noticed that the admin control panel had a long list of links without any obvious organization scheme, which didn't seem to make it easy to find what you were looking for.

This PR re-organizes the links in the admin control panel by breaking the list into sections and using subheadings to group like things together.

It tries to put more frequently-used things in the upper sections. (I don't know if the things in the lower sections are actually used by today's admins.)

The main heading has changed from "Maintenance Operations" to "Administrator Operations" because I assume that's what "adminops" stands for (unless it should be "options"?) and because "Maintenance" got repurposed as a subheading.

ifdb admin panel

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -840,7 +840,7 @@ $filterTypeMap = array(
     "A" => "Adult Content",
     "R" => "Review Embargo");
 
-pageHeader("Maintenance Operations");
+pageHeader("Administrator Operations");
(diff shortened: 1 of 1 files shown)


dfabulich:
Thanks!

PR #1313, opened 2025-08-14
Label external reviews in recent activity

alice-blue:
In the "recent activity" mod tool, external reviews were displaying in a slightly confusing way. Instead of a username, there would just be a blank.

This PR labels external reviews as "external," and prints the word "Someone" when a username is not available.

Before: 
reviewer name screenshot

After:
recent-activity-screenshot-3


dfabulich:
Thanks! I filed #1318 and #1319 on this… it would be nice to show who actually submitted/edited the off-site review here. 

But, for now, I'll merge this, because it's strictly better than what we have.

Issue #1304, opened 2025-08-02
Rich-text editor and/or markdown support


alyshkalia:
A rich-text editor should be added to the site to make it easier for users to compose reviews. And/or, IFDB should support markdown.

PR #1315, opened 2025-08-18
Adding Markdown parsing with Parsedown to reviews

pieartsy:
I added [Parsedown](https://github.com/erusev/parsedown) and implemented in the review and comment text areas.

I made an optional checkbox to convert from markdown to parsedown, and also updated the formatting hints accordingly.

The gif doesn't display that it does links as well.

GIF showing parsedown
parsed review

This is in response to the request to have a markdown and/or rich text editor on IFDB.

https://github.com/iftechfoundation/ifdb/issues/1304

### DIFF
--- a/www/review
+++ b/www/review
@@ -282,7 +282,7 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST' && isset($_REQUEST['delete'])) {
     echo "<b>$title</b>, by $author<br>"
         . showStars($oldrating)
         . "<b>" . htmlspecialcharx($oldsummary) . "</b><br>"
-        . fixDesc(parsedownMultilineText($oldreview), FixDescSpoiler);
+        . fixDesc(parsedownMultilineText($oldreview), FixDescSpoiler | FixDescDemoteHeadings);
 
     echo "</div><form name=\"confirmdelete\" method=\"post\" "
         . "action=\"review?id=$id&userid=$userid\">"
@@ -461,7 +461,7 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST') {
             .   "by <a href=\"showuser?id=$userid\">$username</a>"
             . ($userloc != "" ? " ($userloc)" : "")
             . "</span><br>"
-            . fixDesc($parsedReview, FixDescSpoiler);
+            . fixDesc($parsedReview, FixDescSpoiler | FixDescDemoteHeadings);
 
         if (trim($tags) != "")
             echo "<p><b>Tags:</b> " . htmlspecialcharx($tags);
--- a/www/reviews.php
+++ b/www/reviews.php
@@ -271,9 +271,9 @@ function showReview($db, $gameid, $rec, $specialNames, $optionFlags = 0)
     $rating = $rec['rating'];
     $summary = htmlspecialcharx($rec['summary']);
     if (strcmp($rec['moddate'], '2025-08-26') < 0) {
-        $review = fixDesc($rec['review'], FixDescSpoiler);
+        $review = fixDesc($rec['review'], FixDescSpoiler | FixDescDemoteHeadings);
     } else {
-        $review = fixDesc(parsedownMultilineText($rec['review']), FixDescSpoiler);
+        $review = fixDesc(parsedownMultilineText($rec['review']), FixDescSpoiler | FixDescDemoteHeadings);
     }
     $publicationdate = $rec['publicationdatefmt'];
     $moddatefmt = $rec['moddatefmt'];
--- a/www/util.php
+++ b/www/util.php
@@ -766,6 +766,7 @@ function spoilerWarningScript()
 define("FixDescSpoiler", 0x0001);
 define("FixDescRSS", 0x0002);
 define("FixDescIfic", 0x0004);
+define("FixDescDemoteHeadings", 0x0008);
 function fixDesc($desc, $specials = 0)
 {
     $foundSpoiler = false;
@@ -793,6 +794,7 @@ function fixDesc($desc, $specials = 0)
               'i', 'b', 'u', 'strong', 'em', 'hr',
               'big', 'small', 'tt', 'sup', 'sub',
               'cite', 'blockquote', 'code', 'pre',
+              'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
               'ul', 'ol', 'li', 'dl', 'dt', 'dd'), 1);
 
     // tags that trigger explicit line control mode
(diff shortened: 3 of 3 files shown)


dfabulich:
I think we’ll at least need some kind of verification step to see how many legacy reviews have been affected. It could be as simple as manually reviewing ~100 reviews. If we don’t run into much trouble, maybe we’ll just merge it in and clean up legacy reviews later. (I bet a bunch of them will be improved, containing untransformed Markdown today.)

dfabulich:
I tried copying and pasting the Markdown sample from [this comment](https://github.com/iftechfoundation/ifdb/pull/1315#issuecomment-3198078537) and it looked incorrect on the `viewgame` page when I published it. The bold and italics worked, but the bullets and numbers were untransformed.

Do you repro that, too? (I think it might have to do with calling the parser after calling `fixDesc`.)

image

(Note that the transformed Markdown looks correct in the preview.)

image

pieartsy:
Interesting yeah, I see that too. Thanks for the links to the other pages, I'll work on those. I'll try putting the parser before `fixDesc` also.

dfabulich:
> I also reverted the comments implementation since i haven't figured out how to modify it in all the places where a comment can show up (including the basic review page..) Any guidance with that would be appreciated--or we can make it a separate PR. I think comments are far less used than reviews anyway.

There are a _lot_ of places in IFDB where we might want Markdown, and I think they could all be done in one or more separate PRs once we sort this one out. (And they would _all_ presumably have to deal with the legacy issue, so, in practice, it might not ever turn out to be worth someone's time to handle *every* item on this list.)

* Comments
    * On reviews
    * On polls
    * On lists
    * On user profiles
* Games
    * Description
    * External-link description
    * Off-site review summaries
    * Estimated play-time notes
    * Game news
* Polls
    * Description
    * Poll votes
* Lists
    * Description
    * List item comment (explanation)
* Competitions
    * Description
    * Division description
    * Competition news
* Admin
    * Site news
    * Send admin email
* User profiles
    * Describe yourself

And many of these also have RSS and/or JSON representations!

I'm exhausted just typing all of this out!! So, for now, let's just focus on reviews, and see how _that_ goes.

dfabulich:
I caught another issue. Simple quotes are being translated into `&quot;`.

Try this:

```
Hello this is the **Parsedown** "implementation". _Test_ formatting!

* Also does bullet points
1. And numbers!

(It didn't parse at first cus I didn't check the checkbox.)

Converted automatically to HTML.
```

Even in the preview, it looks like this:

> Hello this is the **Parsedown** &amp;quot;implementation&amp;quot;. _Test_ formatting!

image

www/newitems.php
```diff
@@ -476,7 +479,7 @@ function showNewItemList($db, $first, $last, $items, $options)
                     echo "$stars ";
 
                 if ($len != 0)
-                    echo "<i>\"$summary\"</i>";
+                    echo $parsedown->line("<i>\"$summary\"</i>");
```
dfabulich, line 482: I have a hunch that we want to use parsedown ASAP after reading the review text from the database, i.e. before we call `summarizeHtml` on line 472 above.

pieartsy:
I think that's actually a `fixDesc` issue not an issue with my implementation! Getting rid of `fixDesc` actually solves the issue. However, that function seems to be necessary for spoiler tags, so that's not a real fix. I tried to comment out various `&quot;` related things in the function like this case on line 1135:

```php
        case '"':
            // convert to &quot;
            $desc = substr_replace($desc, "&quot;", $ofs, 1);
            break;
```
but it doesn't seem to be changing anything....also it might break things that _don't_ have parsedown yet.

www/newitems.php
pieartsy: I tried it but that messed it up in the same way that the reviews pages are messed up.

`472       list($summary, $len, $trunc) = summarizeHtml($parsedown->line($r['review']), 140);`
image

I think the ifdb-written ways of parsing html are interfering with the parsedown if they occur afterward. if i parse the markdown _after_ the ifdb utils then it seems to be fine with quotes (i...think the parsedown is parsing the changed `&quot;` stuff back into normal `"` but not sure) but then the aformentioned things like bullet points don't work. A sticky wicket.....

dfabulich:
I think I fixed it in https://github.com/pieartsy/ifdb/pull/2

The bug was that when `fixDesc` saw a `&quot;`, it was turning it into `&amp;quot;`, but the code intended to leave the `&quot;` as-is.

Parsedown seems to always turn `"` into `&quot;` (and that appears to be unconfigurable…? I tried messing around with `setSafeMode` and `setMarkupEscaped` but it didn't seem to change the behavior with `"` characters), which triggered this bug in `fixDesc`.

pieartsy:
Alright I merged your changes. Weird quirk with Parsedown you noticed! I also added a parsedown for when you want to delete your review and it gives a preview of what you're deleting.

dfabulich:
I cooked a little verifier thingy. https://github.com/pieartsy/ifdb/pull/3

I think the main thing we'd want to do is to adjust how Parsedown handles single line breaks.

If you write this in a review on ifdb.org today:

> this
> is
> a
> test

… it will display and render as four lines. In this Parsedown branch, it will display and render as a single line, "this is a test".

pieartsy:
Oh, I left comments on the PR for the verifier a few days ago! I was waiting for you to reply to those.

pieartsy:
Makes sense to me, I think those are pretty good results all things considered. Thank you for looking through them! Can you write that code re: the DB updates?

www/newitems.php
dfabulich: https://github.com/pieartsy/ifdb/pull/9 seems to work for me.

dfabulich (approved):

dfabulich (changes requested):
Gah, just caught a few more issues:

* Headings. `# heading` currently displays literally as "\<h1>heading\</h1>" I'm not entirely sure what we _want_ to do here, because I don't think I want people to randomly slam in their own `<h1>` headings in the page. Maybe we'd have `fixDesc` allow `<h1>` but optionally renumber it, e.g. `<h1>` would become `<h3>`, `<h2>` would become `<h4>`, `<h4>` and higher would become `<h6>`?
* Horizontal rule. `-------` currently displays literally as `<hr />` … I guess we'd do the same thing we did with `<code>` and just allow `<hr>` in `fixDesc`…? Filed https://github.com/pieartsy/ifdb/pull/14 on this.
* Images. `favicon` displays literally as `favicon`. It'd be great if we could just turn that off entirely, but that's a feature only available in Parsedown 2.0 https://github.com/erusev/parsedown/issues/832 The 2.0 beta was released in 2022, but v2.0 never actually shipped. I don't think we want to let users include images in reviews (and we especially don't want to let users _hotlink_ images in reviews)… so I think for now we'd just ignore this, file it as a bug in IFDB, and fix it someday if/when Parsedown 2.0 comes out…?

pieartsy:
Can we just filter out the <h#> and  tags using fixDesc since it runs after parsedown does?

pieartsy:
and what about `<h1>` `<h2>` etc? i agree that people shouldn't be adding h1s. but people havent tried to do this before because markdown wasn't supported before. so i think it'd be ok to strip out?

PR #1355, opened 2025-12-21
Update the "Why did IFDB recommend these?" explanation 

### LINKED ISSUE #449: Update the "Why did IFDB recommend these?" explanation
Now that we're no longer doing the "similar members" recommendation algorithm, the explanation page when you follow the "Why did IFDB recommend these?" link on the home page should be updated: https://ifdb.org/help-crossrec

ShiningSnail:
I want to contribute to the repository, so this is a small first commit and pull request to test things out.

It fixes [Update the "Why did IFDB recommend these?" explanation](https://github.com/iftechfoundation/ifdb/issues/449), which suggests removing the "similar members" algorithm explanation from the [How does IFDB pick games to recommend?](https://ifdb.org/help-crossrec) popup, since the  "similar members" recommendation system is evidently no longer in use.

The text explaning how it worked was still in the code, though it was hidden with PHP, so no one would actually see it. But I removed it entirely.

### DIFF
--- a/www/help-crossrec
+++ b/www/help-crossrec
@@ -15,12 +15,6 @@ members and review sites like
 target="_blank">Baf's Guide</a>.  We're not being paid to promote one
 game over another.
 
-<?php
-$overloaded = true;
-
-if ($overloaded) {
-    ?>
-
 <p>The system picks front-page recommendations by randomly selecting
 a few games with the highest ratings, excluding games that you've
 told us you've already played.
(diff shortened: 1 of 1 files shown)


alice-blue (commented):
This looks good to me, but if we are permanently getting rid of "overloaded," we may also want to delete it from the ifdb-recommends page. See https://github.com/iftechfoundation/ifdb/pull/274/changes

On the other hand, there's an open pull request (https://github.com/iftechfoundation/ifdb/pull/1310) affecting the components of the home page, so I'm not sure of the best timing for making changes to the ifdb-recommends page.  Maybe that could be added as an issue for the future.

alice-blue (approved):

PR #1363, opened 2026-03-10
Adding sandbox to reviews from the author

brirush84:
This copies the sandbox code from the regular reviews page into the special reviews section. It checks the user's sandbox and outputs either (0) or (0,1), then checks the review writer's sandbox and only reveals it if it's in that list.


dfabulich:
Thanks so much for jumping on this!

PR #1367, opened 2026-03-15
Manage suspicious domains

### LINKED ISSUE #1365: Allow IFDB admins to manage the list of suspicious domains
We have a list of suspicious domains checked into `www/newuser`

That means that IFDB admins/moderators can't update/manage that domain list themselves; it requires filing a PR, merging the PR, and deploying the PR (in practice, requiring me personally) to update it.

That sucks, because when we're having an influx of spam accounts from a specific domain, unless I'm very available, we can't respond quickly, forcing us to manually cleanup a bunch of users after the fact.

Instead, we should have an admin tool where admins can add/remove suspicious domains.

(There are actually three lists: one list of emails that will just add "Suspicious domain: yes" to moderator emails, another that will force new users into a "Pending review" state, and another that will simply auto-reject new users from that domain, without emailing moderators.)

https://github.com/iftechfoundation/ifdb/blob/f79282dc7b486da4a73e04ed6ee7a1a7a7ef5fab/www/newuser#L376-L405

alice-blue:
This is still a work in progress. Fixes #1365.

### DIFF
--- a/www/adminops
+++ b/www/adminops
@@ -468,6 +468,142 @@ if (isset($_REQUEST['recentactivity'])) {
              
 //----------------- end of recent activity -------------
 
+//----------------- BEGIN MANAGE SUSPICIOUS DOMAINS -----------
+if (isset($_REQUEST['domains'])) {
+    $new_domain = $_POST['new_domain'] ?? null;
+    $new_suspicion_level = "";
+    if ($_POST['new_suspicion_level'] == 0) {
+        // Change 0 to a string so that conditions
+        // like "if (!$new_suspicion_level)" will work
+        $new_suspicion_level = "zero";
+    } else {
+        $new_suspicion_level = $_POST['new_suspicion_level'] ?? null;
+    }
+    $new_admin_note = $_POST['new_admin_note'] ?? null;
+    $submit = $_POST['submit'] ?? null;
+    // Collect the suspicious domains from the database
+    $result = mysqli_execute_query($db,
+       'SELECT s.domain, s.suspicion_level, s.mod_date, s.modified_by, s.admin_note, u.id, u.name AS user_name
+        FROM suspicious_domains AS s
+        LEFT OUTER JOIN users AS u ON s.modified_by = u.id
+        WHERE (domain, mod_date) IN 
+           (SELECT domain, MAX(mod_date)
+            FROM suspicious_domains
+            GROUP BY domain)
+        ORDER by domain');
+    if (!$result) throw new Exception("Error: " . mysqli_error($db));
+    $latest_domain_rows = mysqli_fetch_all($result, MYSQLI_ASSOC);
+
+    // Start the page
+    pageHeader("Manage Suspicious Domains");
+    echo "<h1>Manage Suspicious Domains</h1>";
+    if (!$new_domain || !$new_suspicion_level) {
+        // User has not entered a domain and suspicion level,
+        // so show the form.
+?>
+        <hr>
+        <h2>Add or update a suspicious domain<h2>        
+        <form action="/adminops?domains" method="post">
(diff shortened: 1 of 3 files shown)


www/newuser
```diff
@@ -381,28 +423,39 @@ function handlePost()
...
+//        . "|rediffmail|renook|i-s-i|smoke911|greenfroginteractive"
+//        . "|thes.ttct.edu.tw"
+//        . ")\b/i",
+//        $email))
         $domainBanned = true;
 
     // Check for TRULY banned domains.  Certain domains are basically
     // 100% spammers, so we can just ignore them out of hand.
-    if (preg_match("/@.*\b(bigmir)\b/i", $email)
+    if (preg_match("/@.*\b($high_suspicion_domains)\b/i", $email)
```
alice-blue: I'm not sure if this matching will still work correctly if there are not any domains in the database in a given category.

dfabulich: Instead of querying for all the domains and then regex matching them, we should extract the domain from the email address and then only query for that domain in `suspicious_domains`.

www/adminops
```diff
@@ -468,6 +468,142 @@ if (isset($_REQUEST['recentactivity'])) {
...
+        // Check to see if the new values make sense. If not,
+        // block the update.
+        $block_domain_update = false;
+        $new_domain_already_listed = false;
+        foreach ($latest_domain_rows as $row) {
+            $domain = htmlspecialchars($row['domain']);
+            $suspicion_level = htmlspecialchars($row['suspicion_level']);
+            if ($domain == $new_domain) { 
+                $new_domain_already_listed = true; 
+                if ($suspicion_level == $new_suspicion_level) {
```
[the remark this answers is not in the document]
dfabulich: > Do you anticipate wanting to add any other trust levels in the future? If so, maybe integers aren't the way to go? It's hard to add e.g. a new level between 2 and 3.

I can't imagine any other trust levels, personally. (And, truth be told, it's _not_ that hard to add them; we'd just do a migration, updating all the 3s to 4s.)

> Are you talking about a moderator log only for activity on suspicious domains, or a moderator log that accounts for all kind of moderator actions?

Either of those is fine, or, as we said on the forum, we could just not log moderator changes at all. We've now written thousands of words about this… I think logging moderator actions is adding a lot of complexity to what would otherwise be a simple feature.

If the table looks like this:

```
CREATE TABLE `suspicious_domains` ( 
  `suspicious_domain_id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `domain` VARCHAR(255) NOT NULL,
  `suspicion_level` tinyint(1) UNSIGNED NOT NULL,
  PRIMARY KEY (suspicious_domain_id),
  UNIQUE KEY `domain` (`domain`)
);
```

Then 90%+ of the questions you're asking go away. We'd mark a domain as unsuspicious by deleting the domain row from the table.

All the remaining questions are then:
* How and where would we log updates? And the answer is "nobody knows" and, for now, "nobody cares"
* Should we record notes/reasons at all? If we do, do we need to keep them private from the IF Archive scrubber? I think the answer is "Don't bother adding notes for now. We can add notes later, and scrub them whenever we get around to it."

I think trying to encode a log of suspicious-domains moderator actions in the `suspicious_domains` table itself (e.g. multiple rows per domain, where each row is really an update) is unwise, precisely because it raises confusing questions, questions that ideally we wouldn't have to discuss or answer at all. It also makes the table more confusing for an outsider to understand. We'd have to add documentation explaining how the whole thing works, and whoever wants to touch the code would have to find that documentation, read it, and understand it.

I think it might have been tempting to define the table as a log table, because you can always _compute_ the current state of the table from the log, but you can't compute a log from the current state of the table. But log tables are complicated and confusing (as you've encountered), whereas ordinary tables with current state are pretty easy to work with.

dfabulich:
We did #1369 instead.

PR #1375, opened 2026-05-27
Add alt text to the small cover art on the viewgame page and full-sized cover art.

### LINKED ISSUE #1378: On viewgame page, lack of alt text on small cover art can be an issue for screen reader users
[Webaim.org](https://webaim.org/techniques/alttext/#functional) says “When an image is the only content inside a link or button, alt text is all that a screen reader has to go on. If it’s empty or missing, screen readers might read the image filename or the URL of the linked page in hopes that this may be helpful to the user, but this is not always the case.”

An IFDB user [says](https://intfiction.org/t/poll-alt-text-for-cover-art-on-ifdb/80337/11):

> Yeah, that’s true in my experience. I think that’s what happens for the main cover image on the ifdb page. For Lost Pig, for example, immediately before the level 1 heading I get this: viewgame coverart&id=mohwfk47yâ¦"
Which is presumably the link to the page with more image info.

Text would be an improvement over reading aloud a URL, and [voters are in favor of identifying the image and telling what the link leads to](https://intfiction.org/t/poll-alt-text-for-cover-art-on-ifdb/80337), so we should put alt text there instead of leaving the alt text blank.

alice-blue:
Based on discussion elsewhere, this PR does the following:

* On the viewgame page, add alt text to the cover art image saying "Cover art for Game Title. Click for full-sized image and copyright information."
* Add an alt text parameter to the coverArtThumbnail function in util.php
* On the large version of the cover art (visible when you click the small cover art on the viewgame page), add the alt text "Cover art for Game Title." (I have not been able to test this, because the dev environment seems to redirect me to the live IFDB site when I try to look at this page.) 

This PR does not add any custom text that describes what the cover looks like, partly because I'm not sure that's actually possible in this dev environment, but hopefully that could be added in the future to one or both of these spots.

Fixes https://github.com/iftechfoundation/ifdb/issues/1378.

### DIFF
--- a/www/util.php
+++ b/www/util.php
@@ -2834,7 +2834,7 @@ function check_admin_privileges($db, $userid) {
 
 }
 
-function coverArtThumbnail($id, $size, $version, $params = "") {
+function coverArtThumbnail($id, $size, $version, $params = "", $alt_text = "") {
     $thumbnail = "/coverart?id=$id&thumbnail=";
     $x15 = round($size * 3 / 2);
     $x2 = $size * 2;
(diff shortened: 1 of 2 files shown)


alice-blue:
I'm going to withdraw this PR for now, because I'm working on an alternative one that also includes custom alt text for images.

Issue #1366, opened 2026-03-14
Duplicate search results when searching by format


dfabulich:
Here's a search for `format:Z*`, which returns multiple instances of Lost Pig, Gostak, Savoir Faire, etc.

https://ifdb.org/search?searchfor=format%3AZ*&searchgo=Search+Games

<img width="303" height="895" alt="Image" src="https://github.com/user-attachments/assets/5859eb23-8f58-4071-a38d-cb6e4e11723a" />

Issue #761, opened 2021-02-18
Rating is too slow! Let me rate any game in a list on any page


qdacsvx:
Each year I upload my If comp ballot. Each year it takes longer!

It takes over an hour because I must open 50+ pages to vote for 50+ games. But what if I could rate each game in the list on IFDB's if comp page? Then it would only take 2 minutes. A huge timesaver.

salty-horse:
@qdacsvx Are you asking ratings controls on both the competition page, and the search page? Maybe pick one of them to be the most useful? I think the comp page one is more relevant to mass-ratings.

Since the star control UI takes up some room, and may clutter the page, is it OK if it's toggled by a checkbox?

dfabulich:
Here's how I think about rating games from the search page.

First, I started with the question: how could we encourage more reviews? "Top reviewers in the last N weeks/months" is part of the answer. #776 

But also, there's the "[IFDB Recommends](https://ifdb.org/search?searchbar=played%3Ano+willplay%3Ano+wontplay%3Ano+reviewed%3Ano+rated%3Ano)" search, linked off of the "IFDB Recommends" section on the home page.

> played:no willplay:no wontplay:no reviewed:no rated:no

Speaking for myself, there are a bunch of interesting games on that list. But I have, in fact, _played_ a number of those games; I just forgot to tell IFDB about it. And for a bunch of them, I haven't played them because I'm not interested.

So what I think I want is a way to quickly skim through that page, checking off games as needed.

Similarly, there's the new "[Unreviewed](https://ifdb.org/search?sortby=lnew&searchfor=%23reviews%3A0+wontplay%3Ano)" link from the "Games" section. I had to manually go in and click "Not interested" on games that are in languages that I don't read/understand. It would be nice to be able to do that right from the search results page.

salty-horse:
OK, I'll tackle this, but first I want to change the game listing in the search page to not use tables.

Do you think there's to put the rating control to the right of each game entry? If not, it might take more vertical space. Should it be toggle-able, to keep the page uncluttered unless you want to review stuff?

dfabulich:
Maybe the answer might be a "display mode switcher" at the top of game search results. Maybe a "compact" mode with no cover art, cramming title, author, year, + star ratings in one span (which could wrap), today's default mode, and a "detail mode" with a description snippet + the entire "have you played" box in each search result…?

salty-horse:
[...]

PR #1383, opened 2026-06-01
Show rating boxes in search page

salty-horse:
This addresses #761.

I added a "show rating boxes" checkbox to the top of the search page.
In the background, the page always loads with all the boxes, but they're hidden. Clicking the checkbox shows/hides them, and updates the address bar and the pagination links to include/remove the new query parameter.

https://github.com/user-attachments/assets/2ba5f4b9-38df-4d38-aeea-3cb85cdee495

I think I tested all the features in the rating box (not shown in video: setting playing times), and also made sure the regular game page still works.

Open issues:
1. The boxes are currently not styled correctly. They use the same style as the sidebar on the game page. They probably shouldn't expand their width - should they be on their own column? Should they be more compact?
2. I used a flexbox to put the checkbox on the same row as the sorting dropdown, but this causes the box to grow in height. I don't know how to fix that :( When it's not a flexbox it's fine. Is this because of the margins set on the drop-down?
3. There's a small quirk with the pagination that probably only happens during local debugging. For my test I used the "highest rated" game list. But as I rate games, this changes their standing in the list, so a page refresh will cause entries to disappear, because they move up/down the list, and are now maybe on a different page. Something to be aware of when you're testing this.
4. To support having multiple boxes on the same page, I added a prefix/suffix of the game ID to elements that needed it. I wasn't super consistent with it, and I think it looks odd to have suffixes that are game IDs. All items of the same kind should start with the same prefix - I will fix it later. Also, if the search results return multiple copies of the same game, like [this bug #1366](https://github.com/iftechfoundation/ifdb/issues/1366), it can cause problems. Should I test for this case?
5. Instead of rendering the HTML for all the boxes and hiding them until the user selects the checkbox, I can make it so selecting it triggers a full page refresh. Technically, this can help with the page size and loading times. Is it a big issue? There aren't too many items per page.
6. How should it look on mobile? ATM it reverts to how the game page presents the box, which is just another section in the game page without a color background.
7. Wasn't tested with Javascript disabled.

### DIFF
--- a/www/editgame
+++ b/www/editgame
@@ -1736,8 +1736,7 @@ function onComboSelectReviewSite(val)
                   <input id="extrevSourceUrl" type=text size=80><br><br>
 
                   <b>Rating:</b> <?php
-                      echo showStarCtl("extrevStarCtl", 0,
-                                       "setExtRevRating", "null");
+                      echo showStarCtl("extrevStarCtl", 0, "setExtRevRating");
                   ?> <br>
                   <input type=hidden name=extrevRating id=extrevRating value=0>
                   <span class=details>If the site uses a different rating
--- a/www/editgame
+++ b/www/editgame
@@ -521,7 +521,6 @@ if ($errMsg) {
                    scriptSrc('/gridform.js'),
                    false);
 
-        initStarControls();
         imageUploadScripts();
 ?>
 
--- a/www/editgame
+++ b/www/editgame
@@ -1036,7 +1036,7 @@ function saveReviewPopup()
             rowDisp = "";
     }
 }
-function setExtRevRating(rating)
+function setExtRevRating(gameId, rating)
 {
     document.getElementById("extrevRating").value = rating;
 }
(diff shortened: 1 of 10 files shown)


alice-blue:
> Should they be more compact?

If there is a way to not show the "Have you played this game" heading on the widget, and instead put "Rate it" at the top of the box, that might save a little vertical space.

salty-horse:
I added a custom CSS class so it can be styled separately than the game page, and removed the title, and the estimated time title and tips.

I think the hours/minutes boxes can be on the same line as "Estimated play time:" and have smaller margins (the margins can also be decreased in the regular game page)
search_rating_box

I'm really bothered by an already-rated box saying "Rate it:" instead of your "Your rating". It was done so the stars don't reflow as you press them but I think I can solve it. I'll submit a separate patch.

salty-horse:
Fixed a bug with updating the "played" flag when submitting time, added a width to the rating box so it's like in the sidebar, removed some spacing between time form lines.



When the game box is larger than the rating box, the rating box stretches. Its contents should probably be aligned to the top.


Should the boxes be aligned to the right? (Example uses `margin-left: auto !important;
  margin-right: 100px !important;`).

aligned_1

Note how the box is squished when the title is longer, even though they are not close.
aligned_2

alice-blue:
> Should the boxes be aligned to the right?

> Note how the box is squished when the title is longer, even though they are not close.

Is there a way to make the game boxes all the same width so that the rating boxes all line up with each other?

Or if the tip boxes are in the way, maybe those could be hidden when ratings boxes are visible.

salty-horse:
I'll hide the tip boxes, and try to fix the difference in width when there's a long game title.

